PR #11, opened 2021-04-13
Add parameter for QuickSight identity region

AWSbrendmb:
The customer may have configured their QuickSight deployment in a different region than where they want to deploy the dashboards.  Added an additional "optional" parameter that uses the just selected region as a default, but offers the customer an option to select an alternate region.

This is to avoid the error on "aws quicksight list-users":
  An error occurred (AccessDeniedException) when calling the ListUsers operation: Operation is being called from endpoint {dashboard region, e.g. us-west-2}, but your identity region is {identity region, e.g. us-east-1}. Please use the {identity region, e.g. us-east-1} endpoint.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cudos/shell-script/lib/common.sh
+++ b/cudos/shell-script/lib/common.sh
@@ -115,12 +115,19 @@ function get_cur_region() {
     export region=${aws_region}
     export AWS_DEFAULT_REGION=${aws_region}
   fi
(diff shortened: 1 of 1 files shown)


yprikhodko (approved):

darken99:
Thank you, @AWSbrendmb

PR #77, opened 2021-10-05
Added ability to run customer-cudos.sh script in a member account, su…

tarfahachem-nbcuni:
…ch as in a dedicated FinOps account. For organizations that prefer to keep their master accounts free of workfloads.

*Issue #, if available:*

*Description of changes: customer-cudos.sh takes optional parameters to be able to run the map function from a member account. Requires a cross-account role to be set up in the root account that the user running the script can assume.


darken99 (changes requested):
Thank you for your efforts.
There are several changes required before we can merge this:

1. Replace `jq` with native bash functions, we do not rely on any 3rd party software
2. Remove hard-coded profile name `cudos`
3. Saving current state of AWS_ environment variables is not acceptable as people can use profiles and those variables will be empty in such a case. There should be another way

darken99:
Legacy automation is deprecated

PR #85, opened 2021-10-12
fixed a few broken commands to get the update template process pass

kentnsw:
Issue #, if available:
n/a

Description of changes:
fixed a few broken commands to get the update template process pass

### DIFF
--- a/tao/shell-script/lib/common.sh
+++ b/tao/shell-script/lib/common.sh
@@ -43,6 +43,9 @@ function config() {
   if [ -z "${region}" ]; then
     get_region
   fi
+  if [ -z "${aws_identity_region}" ]; then
+    get_qs_identity_region
+  fi
   if [ -z "${user_arn}" ]; then
     get_user_arn
   fi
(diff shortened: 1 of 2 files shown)


tao/shell-script/lib/common.sh
```diff
@@ -238,14 +238,14 @@ function update() {
   echo "latest available template version is ${latest_template_version}"
 
   echo -n "Getting currently deployed source version..."
-  current_dashboard_source_version=$(aws quicksight describe-dashboard --dashboard-id $dashboardId --query 'Dashboard.Version.SourceEntityArn' \
-      --aws-account-id $account | cut -f 6 -d \: | cut -f 4 -d \/)
+  current_dashboard_source_version=$(aws quicksight describe-dashboard --dashboard-id $dashboardId --query 'Dashboard.Version.VersionNumber' \
+      --aws-account-id $account)
```
darken99, line 242: This can't be changed like this. The line checks for the latest template version but you're changing it to dashboard version, which is not the same. Please remove this and i'll approve the PR

darken99 (changes requested):

tao/shell-script/lib/common.sh
darken99: Some time ago QuickSight changed the output for the `describe-dashboard` command and there could be two cases, with and without template version. Here is how it looks at my end, please check other dashboards if you have any.
```
aws quicksight describe-dashboard --dashboard-id cudos --aws-account-id XXXXXXXX --query 'Dashboard.Version.SourceEntityArn' --region us-east-1
"arn:aws:quicksight:us-east-1:XXXXXXXX:template/cudos_dashboard_v3/version/120"
```

PR #102, opened 2021-11-16
Python Automation

darken99:
Python beta package


megareez:
we are missing the CFN and parquet file, can that be mitigated?

darken99:
> we are missing the CFN and parquet file, can that be mitigated?

done

megareez (approved):
Yes

megareez:
Epic merge!

PR #217, opened 2022-04-02
Params from cli

iakov-aws:
*Issue #, if available:*

*Description of changes:*


cid/utils.py
```diff
@@ -50,3 +53,55 @@ def get_boto_client(service_name, **kwargs):
...
+    :param default: a default text template
+    :param none_as_disabled: if True and choices is a dict, all choices with None as a value will be disabled
+    :param template_variables: a dict with varibles for template
+    :param break_on_ctrl_c: if True, exit() if user pressed CTRL+C
+
+    :returns: a value choosed by user or provided in command line    
+    """
+    if param_name in params:
+        value = params[param_name]
+        print(f'Using {param_name}={value}, from parameters')
```
darken99, line 72: Should be logger.info()

cid/utils.py
```diff
@@ -50,3 +53,55 @@ def get_boto_client(service_name, **kwargs):
...
+                print(key, value)
+                _choices.append(
+                    questionary.Choice(
+                        title=key,
+                        value=value,
+                        disabled=True if (none_as_disabled and value is None) else False,
+                    )
+                )
+                choices = _choices
+        print()
```
darken99, line 88: Needs to be removed

cid/utils.py
```diff
@@ -50,3 +53,55 @@ def get_boto_client(service_name, **kwargs):
...
+        result = questionary.select(
+            message=f'[{param_name}] {message}:',
+            choices=choices,
+            default=default,
+        ).ask()
+    else: # it is a text entry
+        if default:
+            default=default.format(**template_variables)
+        print()
```
darken99, line 97: Needs to be removed

cid/utils.py
```diff
@@ -50,3 +53,55 @@ def get_boto_client(service_name, **kwargs):
...
+        result = questionary.text(
+            message=f'[{param_name}] {message}:' ,
+            default=default or '',
+        ).ask()
+        if result:
+            result = result.format(**template_variables)
+    if (break_on_ctrl_c and result is None):
+        exit(1)
+    print(f"(Use --{param_name} '{result}' next time you run this)")
```
darken99: Should be logger.info()

cid/helpers/quicksight.py
```diff
@@ -403,21 +400,21 @@ def select_dashboard(self, force=False) -> str:
...
+                choices[key] = None
+            else:
+                choices[key] = dashboard.id
         try:
-            dashboard_id = questionary.select(
-                "Please select installation(s) from the list",
-                choices=selection
-            ).ask()
+            dashboard_id = get_parameter(
+                param_name='dashboard_id',
```
darken99: Use dashboard-id as a param name

cid/helpers/quicksight.py
```diff
@@ -193,15 +194,11 @@ def user(self) -> dict:
...
-                for user in userList:
-                    selection.append(
-                        questionary.Choice(
-                            title=f"{user.get('UserName')} ({user.get('Email')}, {user.get('Role')})",
-                            value=user
-                        )
-                    )
-                self._user =  questionary.select("Please select QuickSight user to use", choices=selection).ask()
+                self._user = get_parameter(
+                    param_name='quicksight_user',
```
darken99: Use quicksight-user as a param name

cid/common.py
```diff
@@ -209,26 +209,21 @@ def track(self, action, dashboard_id):
...
-            selected_dashboard = questionary.select(
-                "Please select dashboard to install",
-                choices=selection
-            ).ask()
-        except Exception as e:
-            logger.debug(e, stack_info=True)
-            print(f'\nEnd: {e}\n')
-            return
+        selected_dashboard = get_parameter(
+            param_name='dashboard-ressource-id',
```
darken99, line 213: Can we use the same dashboard-id here?

cid/common.py
```diff
@@ -483,10 +478,12 @@ def create_datasets(self, _datasets: list) -> dict:
                 print(f'\tfound: {k}', end='')
                 if len(v.keys()) > 1:
                     # Multiple datasets
-                    selected = questionary.select(
-                        f'Multiple "{k}" datasets detected, please select one',
-                        choices=v.keys()
-                    ).ask()
+                    print()
```
darken99, line 481: Why it is here?

cid/helpers/account_map.py
```diff
@@ -217,29 +217,27 @@ def select_metadata_collection_method(self) -> str:
...
-                )
-        selected_source=questionary.select(
-            "Please select account metadata collection method",
-            choices=selection
-        ).ask()
-        if selected_source in account_map_sources.keys():
-            logger.info(f'Selected {selected_source}')
-            self._metadata_source = selected_source
+        selected_source = get_parameter(
+            param_name='account_map_source',
```
darken99: Let's use account-map-source

cid/helpers/account_map.py
```diff
@@ -217,29 +217,27 @@ def select_metadata_collection_method(self) -> str:
...
+        logger.info(f'Selected {selected_source}')
+        self._metadata_source = selected_source
 
         # Collect account list from different sources of user choice
         if self._metadata_source == 'csv':
             finished = False
             while not finished:
-                mapping_file = click.prompt("Enter file path", type=str)
+                mapping_file = get_parameter(
+                    param_name='account_map_file',
```
darken99: Same as above for using dashes instead of underscores

cid/helpers/athena.py
```diff
@@ -50,10 +50,12 @@ def CatalogName(self) -> str:
                 default_catalog = [d for d in glue_data_catalogs if d['CatalogName'] == self.defaults.get('CatalogName')]
                 if not len(default_catalog):
                     # Ask user
-                    self._CatalogName = questionary.select(
-                        "Select AWS DataCatalog to use",
+
+                    self._CatalogName = get_parameter(
+                        param_name = 'glue_data_catalog',
```
darken99, line 55: Same as above for using dashes instead of underscores

cid/helpers/athena.py
```diff
@@ -82,10 +84,11 @@ def DatabaseName(self) -> str:
                     self._DatabaseName = default_database.pop().get('Name')
                 else:
                     # Ask user
-                    self._DatabaseName = questionary.select(
-                        "Select AWS Athena database to use",
-                        choices=[d['Name'] for d in athena_databases]
-                    ).ask()
+                    self._DatabaseName = get_parameter(
+                        param_name = 'athena_database',
```
darken99, line 88: Same as above for using dashes instead of underscores

darken99 (changes requested):

cid/cli.py
```diff
@@ -18,70 +32,102 @@
...
+    app = ctx.obj
+    app.open(dashboard_id=kwargs.get('dashboard_id'))
 
 
-@main.command()
-@click.pass_obj
-def cleanup(App: Cid):
+@main.command(**command_params)
+@click.pass_context
+def cleanup(ctx: Cid):
```
darken99, line 126: ctx is not an instance of Cid

darken99 (changes requested):

cid/utils.py
iakov-aws: yes.

cid/utils.py
iakov-aws: the issue is that we have print(..., end='') in several places, so the prompt will be in the same line. We need to remove  end='' or leave the print here

cid/utils.py
iakov-aws: same as above

cid/utils.py
iakov-aws: info is not visble by the user. It is a message to the user. We need to make info visible or leave it as print. I am also thinking about printing the full parameters list at the end, so that user can save this for the next time

cid/helpers/quicksight.py
iakov-aws: yes

cid/common.py
iakov-aws: indeed needs to be remvoed, as this print is moved to the get_parameters function. was initally due to end="" in prints

cid/helpers/account_map.py
iakov-aws: yes

cid/helpers/account_map.py
iakov-aws: yes

cid/helpers/athena.py
iakov-aws: yes

cid/helpers/athena.py
iakov-aws: yes

cid/cli.py
iakov-aws: fixed

iakov-aws:
Thanks for comments. 

I will also need to refactor to pass params to the code of functions directly, not via cli. This will help in lambda custom function.

cid/utils.py
darken99: No, this should be covered in documentation but not printed to stdout

cid/common.py
iakov-aws: well, dashboard-id is used when it s a quicksight dashboard-id in customer's acc. Here it is the reference to dashboard in ressource.yaml this is why i make this difference. Otherwise user will put "cudos" instead of "CUDOS" etc see ressource.yaml keys vs names vs dashboard-id.

cid/utils.py
iakov-aws: let's have a call next week.

cid/common.py
iakov-aws: This commit changes slightly the logic of dashboard choice. Before it was selected by the keys in dashboard section of ressource.yaml, now dashboard-id is used

https://github.com/aws-samples/aws-cudos-framework-deployment/pull/217/commits/5372e8d5ba3280a53be3360ea448c61306cfd09b

darken99 (approved):

PR #222, opened 2022-04-06
Adding support for QuickSight regions other than us-east-1

GeorgeRZKS:
*Issue #, if available:*

The cdi-cli lacks support for QuickSight in regions other than us-east-1

*Description of changes:*

Added support of an additional command line parameter: --quicksight_region


darken99 (changes requested):
QuickSight working region and Identity region are different instances. You can't simply change a region when creating a QuickSight client object

darken99:
Implemented in #231

PR #252, opened 2022-05-10
Bugfix: resource management

darken99:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -920,7 +920,7 @@ class Cid:
 
         # create missing views
         if len(missing_views):
-            print(f"Missing views: {', '.joinmissing_views}")
+            print(f"Missing views: {', '.join(missing_views)}")
             for view_name in missing_views:
                 self.create_or_update_view(view_name, recursive=recursive, update=update)
 
--- a/cid/common.py
+++ b/cid/common.py
@@ -4,6 +4,7 @@ import json
 import logging
 from pathlib import Path
 from string import Template
+from typing import Dict
 from pkg_resources import resource_string
 
 import click
@@ -13,7 +14,7 @@ from botocore.exceptions import ClientError, NoCredentialsError, CredentialRetri
 
 from cid import utils
 from cid.plugin import Plugin
-from cid.utils import get_parameter
+from cid.utils import get_parameter, unset_parameter
 from cid.helpers.account_map import AccountMap
 from cid.helpers import Athena, CUR, Glue, QuickSight, Dataset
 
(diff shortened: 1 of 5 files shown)


cid/helpers/quicksight/__init__.py
```diff
@@ -341,7 +352,7 @@ def list_data_sources(self) -> list:
                 return result.get('DataSources')
         except self.client.exceptions.AccessDeniedException:
             logger.info('Access denied listing data sources')
-            return list()
+            raise
         except Exception as e:
             logger.debug(e, stack_info=True)
             return list()
```
iakov-aws, line 358: Would it be more logical to raise here as well?

cid/common.py
```diff
@@ -924,7 +920,7 @@ def create_or_update_dataset(self, dataset_definition: dict, arn: str=None,recur
 
         # create missing views
         if len(missing_views):
-            print(f'\tMissing Athena views: {missing_views}')
+            print(f"Missing views: {', '.joinmissing_views}")
```
iakov-aws, line 923: joinmissing_views undefined

iakov-aws (changes requested):

cid/helpers/quicksight/__init__.py
darken99: Maybe, but will requires to change in other places as well

cid/common.py
```diff
@@ -896,14 +890,16 @@ def create_or_update_dataset(self, dataset_definition: dict, arn: str=None,recur
         # else user will be suggested to choose database
 
         columns_tpl = {
-            'athena_datasource_arn': next(iter(self.qs.athena_datasources)),
+            'athena_datasource_arn': next(iter(v.arn for v in self.qs.athena_datasources.values())),
```
iakov-aws, line 893: can we add a fallback here? 
```
'athena_datasource_arn': next(iter(v.arn for v in self.qs.athena_datasources.values()), None),
```

iakov-aws (approved):

cid/helpers/quicksight/__init__.py
```diff
@@ -572,11 +583,14 @@ def describe_dataset(self, id, timeout: int=1) -> dict:
 
         return self._datasets.get(id, None)
 
-    def discover_datasets(self):
+    def discover_datasets(self, _datasets: list=None):
```
iakov-aws, line 586: suggest following

```python
  def discover_datasets(self, datasets: list=None):
        """ Discover datasets in the account 
        :datasets : list of datast ids, if not provided or empty, will discover all
        """

        logger.info('Discovering datasets')
        self._datasets =  self._datasets or {}

        try:
            dataset_ids = datasets or [d.get('DataSetId') for d in self.list_data_sets()]:
            for dataset in dataset_ids:
                try:
                    self.describe_dataset(dataset.get('DataSetId'))
                except Exception as e:
                    logger.debug(e, stack_info=True)
                    continue
        except self.client.exceptions.AccessDeniedException:
            logger.info('Access denied listing datasets')
        except Exception as e:
            logger.debug(e, stack_info=True)
            logger.info('No datasets found')
```

PR #262, opened 2022-05-18
improve dataset management

iakov-aws:
*Issue #, if available:*
improve dataset management: avoid errors if workgroup not found and add explicit log when dataset not found 
*Description of changes:*

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -261,8 +261,12 @@ class Cid:
             dashboard_definition.update({'datasets': {}})
         for dataset_name in required_datasets:
             arn = next((v.arn for v in self.qs._datasets.values() if v.name == dataset_name), None)
+
             if arn:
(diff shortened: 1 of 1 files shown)


cid/common.py
```diff
@@ -261,8 +261,12 @@ def deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs)
             dashboard_definition.update({'datasets': {}})
         for dataset_name in required_datasets:
             arn = next((v.arn for v in self.qs._datasets.values() if v.name == dataset_name), None)
+
             if arn:
                 dashboard_definition.get('datasets').update({dataset_name: arn})
+            else:
```
darken99, line 267: Let's use get_datasets() here and make dashboard datasets a priority over other

darken99 (changes requested):

PR #263, opened 2022-05-18
fix QuickSight user choice

iakov-aws:
*Issue #, if available:*

*Description of changes:*
Fix for cases when we do not have permission to read user in quicksight.

### DIFF
--- a/cid/helpers/quicksight/__init__.py
+++ b/cid/helpers/quicksight/__init__.py
@@ -81,6 +81,9 @@ class QuickSight():
             if not self._user:
                 # If no user match, ask
                 self._user = self.select_user()
+            if not self._user:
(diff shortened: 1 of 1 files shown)


cid/helpers/quicksight/__init__.py
```diff
@@ -81,6 +81,9 @@ def user(self) -> dict:
             if not self._user:
                 # If no user match, ask
                 self._user = self.select_user()
+            if not self._user:
```
darken99, line 84: We have to detect the username from the session

darken99 (changes requested):

cid/helpers/quicksight/__init__.py
iakov-aws: yes. indeed

darken99 (approved):

PR #295, opened 2022-07-10
fix bats actions

iakov-aws:
*Issue #, if available:*

*Description of changes:*
This change adds an actions to install bats instead of broken apt-get install 


iakov-aws:
Need to internalize the code of action

PR #298, opened 2022-07-13
Manage QS subscriptions and Athena Workgroups

iakov-aws:
*Issue #, if available:*

*Description of changes:*


assets/recommended_permissions.json
```diff
@@ -110,6 +112,15 @@
...
               "arn:aws:s3:::<ATHENA-QUERY-RESULTS-BUCKET>/*"
             ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:CreateBucket",
+            ],
+            "Resource": [
+              "arn:aws:s3:::<ATHENA-QUERY-RESULTS-BUCKET>"
```
yprikhodko, line 122: does this makes sense? if it will be created by the code which you've added the name pattern is not defined by customer, how then they would be able to provide access with this policy?

iakov-aws: This is not a real json, it is rather a template

PR #328, opened 2022-08-15
add delete failed datasource

iakov-aws:
*Issue #, if available:*

*Description of changes:*
Avoid keeping all failed datasources + change name of datasource to Athena-CID for better identification. 


cid/helpers/quicksight/__init__.py
```diff
@@ -211,7 +211,7 @@ def create_data_source(self) -> bool:
         params = {
             "AwsAccountId": self.account_id,
             "DataSourceId": str(uuid.uuid4()),
-            "Name": "Athena",
+            "Name": "Athena-CID",
```
darken99, line 214: Why do you think we should rename datasource?

cid/helpers/quicksight/__init__.py
```diff
@@ -235,6 +235,15 @@ def create_data_source(self) -> bool:
...
             if not datasource.is_healthy:
                 logger.error(f'Data source creation failed: {datasource.error_info}')
+                if get_parameter(
+                    param_name='quicksight-delete-failed-datasource',
+                    message=f'Data source creation failed: {datasource.error_info}. Delete?',
+                    choices=['yes', 'no'],
+                ) == 'yes':
+                    try:
+                        self.delete_data_source(datasource.id)
+                    except self.client.exceptions.AccessDeniedException as e:
```
darken99, line 245: this should be handled inside delete_data_source()

cid/helpers/quicksight/__init__.py
```diff
@@ -550,6 +559,16 @@ def delete_dashboard(self, dashboard_id):
...
 
+    def delete_data_source(self, datasource_id):
+        """ Deletes an AWS QuickSight dashboard """
+        paramaters = {
+            'AwsAccountId': self.account_id,
+            'DataSourceId': datasource_id
+        }
+        logger.info(f'Deleting DataSource {datasource_id}')
+        result = self.client.delete_data_source(**paramaters)
+        return result
```
darken99, line 570: We have to also remove it from the internal resources structure if it existed

darken99 (changes requested):

cid/helpers/quicksight/__init__.py
iakov-aws: So that we could distinguish once created by us from all other athena datasource in qs

cid/helpers/quicksight/__init__.py
iakov-aws: ok

iakov-aws: well after some thoughts: how we can know from delete_datasource() if the error is critical or not. We know it only in the context of calling function. So error handling (or ignoring) must be also at this level.

cid/helpers/quicksight/__init__.py
iakov-aws: yes. fixed

cid/helpers/quicksight/__init__.py
darken99: This is not needed, please revert

iakov-aws: not agree but done.

darken99 (approved):

PR #345, opened 2022-08-25
Enhancement: CidBase class

darken99:
*Issue #, if available:*

*Description of changes:*
New Class CidBase to simplify session management in other classes.


cid/common.py
```diff
@@ -23,7 +24,7 @@
 
 logger = logging.getLogger(__name__)
 
-class Cid:
+class Cid(CidCore):
```
iakov-aws, line 27: Cid And Helpers seems too different to derive them from one parent class.

iakov-aws (approved):

PR #350, opened 2022-08-29
Change: Stop reading legacy deployments information

darken99:
*Issue #, if available:*

*Description of changes:*
As we do autodiscovery of the existing deployments, have resource take-over mechanism and share feature we no longer need to workaround reading legacy `.json` files with deployment info.


iakov-aws (approved):
Love that one!
❤️❤️❤️

PR #386, opened 2022-10-17
fix-datasource-management

iakov-aws:
*Issue #, if available:*

*Description of changes:*
Correctly manage quicksight-datasource-id

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -943,79 +943,93 @@ class Cid():
         cur_required = dataset_definition.get('dependsOn', dict()).get('cur')
         athena_datasource = None
 
-        if not len(self.qs.athena_datasources):
+        if get_parameters().get('quicksight-datasource-id'):
+            # We have explicit choice of datasource
+            datasource_id = get_parameters().get('quicksight-datasource-id')
+
+            try:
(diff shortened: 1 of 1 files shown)


cid/common.py
```diff
@@ -943,79 +943,93 @@ def create_or_update_dataset(self, dataset_definition: dict, dataset_id: str=Non
...
 
-        if not len(self.qs.athena_datasources):
+        if get_parameters().get('quicksight-datasource-id'):
+            # We have explicit choice of datasource
+            datasource_id = get_parameters().get('quicksight-datasource-id')
+
+            try:
+                athena_datasource = self.qs.describe_data_source(datasource_id)
+            except self.qs.client.exceptions.AccessDeniedException:
+                athena_datasource = Datasource(raw={
```
darken99, line 953: Why do we use the ID anyway even we don't have access to it?

darken99 (approved):

cid/common.py
iakov-aws: Here user provides explicitly the id. We do not need access to use it. User can just have a describe permission denied

PR #403, opened 2022-11-11
Bugfix: in case if exising dataset or folder the exception was raised fixing that.

iakov-aws:
Also add a better comment for SPICE capacity exception

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/helpers/quicksight/__init__.py
+++ b/cid/helpers/quicksight/__init__.py
@@ -395,9 +395,10 @@ class QuickSight(CidBase):
     def create_folder(self, folder_name: str, **create_parameters) -> dict:
         """Create a new folder"""
         logger.info('Creating QuickSight folder')
+        folder_id = str(uuid.uuid4())
         create_parameters.update({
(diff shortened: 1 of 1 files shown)


cid/helpers/quicksight/__init__.py
```diff
@@ -411,6 +412,7 @@ def create_folder(self, folder_name: str, **create_parameters) -> dict:
             folder = self.describe_folder(result['FolderId'])
             return folder
         except self.client.exceptions.ResourceExistsException:
+            return self.describe_folder(folder_id)
             logger.error('Folder already exists')
```
darken99, line 416: Please change to logger.info()

darken99 (changes requested):

cid/helpers/quicksight/__init__.py
iakov-aws: Must be deleted

darken99 (approved):

PR #411, opened 2022-11-19
fix for cases when template is not detected

iakov-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/helpers/quicksight/dashboard.py
+++ b/cid/helpers/quicksight/dashboard.py
@@ -43,7 +43,7 @@ class Dashboard(CidQsResource):
     @property
     def template_id(self) -> str:
         if isinstance(self.deployedTemplate, CidQsTemplate):
-            return self.deployedTemplate.arn
+            return self.deployedTemplate.id
         return None
 
     @property
@@ -67,6 +67,8 @@ class Dashboard(CidQsResource):
     def latest_version(self) -> int:
         if isinstance(self.sourceTemplate, CidQsTemplate):
             return self.sourceTemplate.version
+        else:
+            return -1
 
     @property
     def health(self) -> bool:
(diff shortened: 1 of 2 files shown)


cid/helpers/quicksight/dashboard.py
```diff
@@ -40,6 +40,18 @@ def deployedTemplate(self) -> CidQsTemplate:
     def deployedTemplate(self, template: CidQsTemplate) -> None:
         self._deployedTemplate = template
 
+    @property
+    def template_id(self) -> str:
+        if isinstance(self.deployedTemplate, CidQsTemplate):
+            return self.deployedTemplate.arn
```
darken99, line 46: Shoudn't we use `self.deployedTemplate.id` here?

cid/helpers/quicksight/dashboard.py
```diff
@@ -53,7 +65,8 @@ def latest(self) -> bool:
 
     @property
     def latest_version(self) -> int:
-        return self.sourceTemplate.version
+        if isinstance(self.sourceTemplate, CidQsTemplate):
+            return self.sourceTemplate.version
```
darken99, line 69: What if not an instance of `CidQsTemplate`? Should we return `None` or `-1`?

darken99 (changes requested):

cid/helpers/quicksight/dashboard.py
iakov-aws: indeed

cid/helpers/quicksight/dashboard.py
iakov-aws: No sure what is the purpose of -1 but we we can do that for consistency

darken99 (approved):

PR #424, opened 2022-12-07
Init cmd

gaborschulz-aws:
*Description of changes:*
New implementation of `init` command


cid/builtin/core/data/glue/table_columns.json
```diff
@@ -0,0 +1,254 @@
+[
```
iakov-aws, line 1: cid/builtin/core/data/glue/table_columns.json -> cid/builtin/core/data/glue/cur_columns.json

?

cid/common.py
```diff
@@ -1279,8 +1279,15 @@ def map(self, **kwargs):
             self.accountMap.create(v)
 
     @command
-    def init(self, **kwargs):
-        """ Initialize account resources for deployment """
-        from commands import InitCommand
-        cmd = InitCommand(cid=self, **kwargs)
+    def initqs(self, **kwargs):
+        """ Initialize QuickSight resources for deployment """
+        from commands import InitQsCommand
```
iakov-aws, line 1284: is there a reason to bring this import to function level. Can lead to errros that are hard to debug

cid/commands/init_cur.py
```diff
@@ -201,103 +199,7 @@ def _create_query_results_bucket(self):
...
-        while not email or '@' not in email or '.' not in email:
-            counter += 1
-            email = get_parameter('qs-notification-email', 'Notification email', default=self.cid.organizations.get_account_email())
-            if '@' not in email or '.' not in email:
-                print(f'\t{email} does not seem to be a valid email. Please, try again.')
-                unset_parameter('qs-notification-email')
-                if counter >= MAX_ITERATIONS:
-                    raise CidCritical()
-        return email           
+            sys.exit(1)
```
iakov-aws, line 202: raise CidCrirtical?

cid/common.py
iakov-aws: also probably you mean

from cid.commands import InitQsCommand

cid/builtin/core/data/glue/table_columns.json
gaborschulz-aws: Renamed to follow our naming convention

cid/common.py
gaborschulz-aws: I'm using type hints and leaving the InitQsCommand as a top level import will lead to circular dependencies

cid/commands/init_cur.py
gaborschulz: Fixed

cid/commands/init_cur.py
```diff
@@ -193,13 +193,10 @@ def _create_query_results_bucket(self):
             print(f'\tAthena S3 Bucket...\tCreated ({self.bucket_name})')
         except CidError as ex:
             print(f'\tAthena S3 Bucket...\t{ex} ({self.bucket_name})')
-        except CidCritical as ex:
-            print(f'\tAthena S3 Bucket...\t{ex} ({self.bucket_name})')
-            sys.exit(1)
         except Exception as ex:  # pylint: disable=broad-except
             print('\tAthena S3 Bucket...\tFailed')
```
iakov-aws, line 197: no need in both print and error

gaborschulz-aws: Removed already :-)

cid/helpers/iam.py
```diff
@@ -178,6 +178,27 @@ def ensure_role_does_not_exist(self, role_name):
         except self.client.exceptions.DeleteConflictException:
             logger.debug(f'Role  {role_name!r} is still used. Skipping')
 
+    def create_role(self, role_name: str, assume_role_policy_document: str) -> str:
+        """Create IAM role with a specified name and role policy document for assuming the role"""
+        response = self.client.create_role(RoleName=role_name, AssumeRolePolicyDocument=assume_role_policy_document, Tags=self.default_tag_list)
```
iakov-aws, line 183: This can fail with many reasons: 
- role can exist and permissions are the main ones

gaborschulz-aws: Good point. I'm letting that exception bubble up and I'm handling it in `_create_glue_crawler_role` in the `InitCurCommand`, because if we want to use this method anywhere else, I would probably handle the exception differently there

cid/helpers/glue.py
```diff
@@ -32,3 +36,54 @@ def delete_table(self, name, catalog, database):
...
+                CatalogId=catalog_id,
+                DatabaseInput={
+                    "Name": name,
+                },
+                Tags=self.default_tag_dict
+            )
+        except self.client.exceptions.AlreadyExistsException as ex:
+            logger.debug(ex, exc_info=True)
+            logger.info("Database %s cannot be created in %s: %s", name, catalog_id, ex)
+            raise CidError() from ex
```
iakov-aws, line 53: In other paces we use a logic of 'ensure' this means if we ask to create it will create or return exisitng. Simiarly with delete if the final state is achieved we do not need to raise a problem. 

Please can you review if this fits?

cid/commands/init_cur.py
```diff
@@ -0,0 +1,228 @@
...
+        custom_policies = {'AWSCURCrawlerComponentFunction': 'aws_cur_crawler_component_function.json', 'AWSCURKMSDecryption': 'aws_cur_kms_decryption.json'}
+
+        print('\tAttaching policies...')
+        for policy in managed_policies:
+            print(f'\t\t{policy}')
+            self.cid.iam.attach_policy(role_name=self.cid_crawler_role_name, policy_arn=policy)
+
+        for name, policy_file in custom_policies.items():
+            policy_document = pkg_resources.read_text('cid.builtin.core.data.iam', policy_file)
+            policy_document = inject_variables(source=policy_document, variables=self.variables)
```
iakov-aws, line 124: Wgt we do not use a template as in other places?

PR #470, opened 2023-01-19
Template without version

tbuatois:
*Issue #, if available:*

*Description of changes:*


cid/helpers/quicksight/__init__.py
```diff
@@ -199,14 +199,21 @@ def discover_dashboard(self, dashboardId: str) -> Dashboard:
...
-                region=_template_arn.split(':')[3],
-                version_number=int(_template_arn.split('/')[-1]),
-            )
-            if isinstance(_template, CidQsTemplate):
-                dashboard.deployedTemplate = _template
+            
+            ## Check for version_number
+            template_version_nb = int(_template_arn.split('/')[-1]) if _template_arn.split('/')[-1].isdecimal() else None
+            
+            if  template_version_nb:
```
iakov-aws, line 206: why not leaving template_version_nb as None?

describe works as expected

yprikhodko: If template_version_nb is None we still need to describe_template

yprikhodko (commented):
@tbuatois , can we assume further in the code that if template version is empty we consider it as default version from resources.yaml?
this would allow us to catch major version update then

cid/helpers/quicksight/__init__.py
iakov-aws: We discussed this and it looks like if version is none we get only the latest that does not make sense for further actions.

tbuatois:
> @tbuatois , can we assume further in the code that if template version is empty we consider it as default version from resources.yaml? this would allow us to catch major version update then

@yprikhodko 
What I've done here is to let the user know that we could not figure out the version of the deployedTemplate and so say : please update normally, if this fail use --recursive option. If there is a major update, the deployment will fail and the user will just need to re run the command. 
We can also directly propose to use the recursive mode in that case.

PR #471, opened 2023-01-21
Export SQL views and add SQL diff

iakov-aws:
*Issue #, if available:*

*Description of changes:*


cid/common.py
```diff
@@ -298,68 +301,104 @@ def _deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs
...
+                    template_id=dashboard_definition.get('templateId'),
+                    account_id=dashboard_definition.get('sourceAccountId'),
+                    region=dashboard_definition.get('region', 'us-east-1')
+                )
+            except CidError as exc:
+                raise CidCritical(exc) # Cannot proceed without a valid template
+            dashboard_definition['sourceTemplate'] = source_template
+            print(f'\nLatest template: {source_template.arn}/version/{source_template.version}')
+            compatible = self.check_dashboard_version_compatibility(dashboard_id)
+        elif dashboard_definition.get('data'):
```
yprikhodko, line 324: could 'data' key doesn't exist? maybe better to change with elif 'data' in dashboard_definition so it doesn't throw an error?

cid/common.py
```diff
@@ -298,68 +301,104 @@ def _deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs
...
-                        unmatched.update({k: {'expected': v, 'found': dataset_fields.get(k)}})
-                if unmatched:
-                    logger.warning(f'Dataset "{dataset_name}" ({ds.id}) is missing required fields. {(unmatched)}')
-                    continue
+        for dataset_name in required_datasets_names:
+            # First try to find the dataset with the id
+            dataset = self.qs.describe_dataset(id=dataset_name)
+            if isinstance(dataset, Dataset):
+                logger.debug(f'Found dataset {dataset_name} with id match = {ds.arn}')
+                dashboard_definition['datasets'][dataset_name] = dataset.arn
```
yprikhodko, line 359: why we are not checking required fields in this closure?

cid/cli.py
```diff
@@ -135,13 +137,17 @@ def export(ctx, **kwargs):
...
-        --output              A filename (.yaml)
+        --analysis-name              Analysis you want to share (not needed if analysis-id is provided).
+        --analysis-id                ID of analysis you want to share (open analysis in browser and copy id from url)
+        --template-id                Template Id
+        --dashboard-id               Target Dashboard Id
+        --template-version           Version description vX.Y.Z
+        --reader-account             Account id with howm you want to share or *
+        --dashboard-export-method
+               (definition|template) A method (definition=pull json definition of Analysis OR template=create QuickSught Teamplate)
+        --export-known-datasets
```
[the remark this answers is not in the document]
iakov-aws: In most cases it will be dashboards created on top of existing. So we do not need to override default behaviour. For example most of datasets are dependent on account_map. We do not want to override account map with a concrete implementation in source account.

cid/common.py
```diff
@@ -1174,13 +1229,49 @@ def create_or_update_dataset(self, dataset_definition: dict, dataset_id: str=Non
...
+            if update_dataset and get_parameters().get('on-drift', 'show').lower() != 'override' and isatty() and not cur_required:
+                while True:
+                    diff = self.qs.dataset_diff(found_dataset.raw, compiled_dataset)
+                    if diff and diff['diff']:
+                        cid_print(f'<BOLD>Found a difference between existing dataset <YELLOW>{found_dataset.name}<END> <BOLD>and the one we want to deploy. <END>')
+                        cid_print(diff['printable'])
+                        choice = get_parameter(
+                            param_name='dataset-' + found_dataset.name.lower().replace(' ', '-') + '-override',
+                            message=f'The existing dataset is different. Override?',
+                            choices=['retry diff', 'proceed and override', 'keep existing', 'stop'],
```
yprikhodko, line 1247: maybe let's rename stop to exit?

cid/common.py
```diff
@@ -1174,13 +1229,49 @@ def create_or_update_dataset(self, dataset_definition: dict, dataset_id: str=Non
...
+                            update_dataset = True
+                            break
+                        elif choice == 'keep existing':
+                            update_dataset = False
+                            break
+                        else:
+                            raise CidCritical(f'User choice is not to update {found_dataset.name}.')
+                    elif not diff:
+                        if not get_parameter(
+                            param_name=found_dataset.name.lower().replace(' ', '-') + '-override',
```
yprikhodko, line 1263: is there any way to get rid of this replace?

cid/common.py
```diff
@@ -1174,13 +1229,49 @@ def create_or_update_dataset(self, dataset_definition: dict, dataset_id: str=Non
...
             if update:
-                self.qs.update_dataset(compiled_dataset)
+                update_dataset = True
             elif found_dataset.name != compiled_dataset.get('Name'):
                 print(f"Dataset found with name {found_dataset.name}, but {compiled_dataset.get('Name')} expected. Updating.")
+                update_dataset = True
+            if update_dataset and get_parameters().get('on-drift', 'show').lower() != 'override' and isatty() and not cur_required:
+                while True:
+                    diff = self.qs.dataset_diff(found_dataset.raw, compiled_dataset)
+                    if diff and diff['diff']:
```
yprikhodko, line 1241: Shouldn't this be part of helper code to keep common.py more high level?

cid/common.py
```diff
@@ -1231,15 +1322,49 @@ def create_or_update_view(self, view_name: str, recursive: bool=True, update: bo
                     self.glue.create_or_update_table(view_name, view_query)
                 else:
                     if 'CREATE OR REPLACE' in view_query.upper():
-                        print(f'Updating view: "{view_name}"')
-                        self.athena.execute_query(view_query)
+                        update_view = False
+                        while get_parameters().get('on-drift', 'show').lower() != 'override' and isatty():
+                            cid_print(f'Analysing view {view_name}')
+                            diff = self.athena.get_view_diff(view_name, view_query)
+                            if diff and diff['diff']:
```
yprikhodko, line 1329: same comment, could it be in the helper?

cid/common.py
iakov-aws: get returns none if data is not in the dict.

cid/common.py
iakov-aws: name can contain spaces. parameter names cannot. i do not see how we can workaround this without a replace

cid/common.py
iakov-aws: I would love to, but there are too many code related to user interaction and to logic. i do not see how it can be pushed outside without raising the complexity.

cid/common.py
iakov-aws: same answer as above :)

cid/export.py
```diff
@@ -96,109 +139,217 @@ def export_analysis(qs):
...
+        # Analyse dependancies: if the dependancy is CUR there is a special flag
+        deps = view_data.get('dependsOn', {})
+        non_cur_dep_views = []
+        for dep_view in deps.get('views', []):
+            if cur_helper.table_is_cur(name=dep_view):
+                logger.debug(f'{dep_view} is cur')
+                view_data['dependsOn']['cur'] = True
+                # replace cur table name with a variable
+                if isinstance(view_data.get('data'), str):
+                    view_data['data'] = view_data['data'].replace(f'{dep_view}', '${cur_table_name}')
```
yprikhodko, line 198: cur tables treated separately as we don't create CUR table, but assume it should be created by customer, right? Could you please add this as a comment

iakov-aws: Added

cid/export.py
```diff
@@ -96,109 +139,217 @@ def export_analysis(qs):
...
+                dep_cur = True
+        datasets[dataset_name] = {
+            'data': dataset_data,
+            'dependsOn': {'views': dependancy_views},
+        }
+        if dep_cur:
+            datasets[dataset_name]['dependsOn']['cur'] = True
 
-        datasets[dataset_arn] = dataset_data
+    all_databases = list(set(all_databases))
```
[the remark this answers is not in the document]
iakov-aws: There is an explaination just bellow

cid/export.py
```diff
@@ -96,109 +139,233 @@ def export_analysis(qs):
...
-            'datasets': [dataset['Name'].replace(' ', '-') for dataset in datasets.values()]
+    dashboard_export_method = None
+    if get_parameters().get('template-id'):
+        dashboard_export_method = 'template'
+    else:
+        dashboard_export_method = get_parameter(
+            'dashboard-export-method',
+            message='Please choose export method (pull a json "definition" or create a "template")',
+            choices=[
+                'definition',
```
yprikhodko, line 261: Suggest to change to:
template definition file
shared template in QuickSight account

cid/helpers/athena.py
```diff
@@ -437,3 +361,86 @@ def delete_view(self, name: str, catalog: str=None, database: str=None):
...
+        tmp_name = 'cid_tmp_deleteme'
+        existing_sql = ''
+        try:
+            existing_sql = self.query(f'SHOW CREATE VIEW {name}', include_header=True)
+            existing_sql = '\n'.join([line[0] for line in existing_sql][1:])
+        except Exception as exc:
+            print(exc)
+            return None
+        try:
+            tmp_sql = re.sub(r'(CREATE OR REPLACE VIEW) (.+?) (AS.*)', r'\1 ' + tmp_name +  r' \3', sql)
```
yprikhodko, line 376: this requires explanation in the comments why we're doing it this way

cid/queries/ahq-queries.json
```diff
@@ -1,12 +0,0 @@
-{
```
yprikhodko, line 1: why these files removed? We still making this checks for customers who deploying dashboards over CUR tables not created by our CFN

iakov-aws: These are not used since a long time. These check are replaced by working with glue metadata objects.

cid/export.py
iakov-aws: not sure to understand what is requested. can you provide full line with your option? will be happy to apply.

iakov-aws: These must be short as the same values will be used in command line parameters

iakov-aws: I can probably rewrite this to achieve the same results

cid/common.py
```diff
@@ -385,7 +385,8 @@ def _deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs
 
                 if not matching_datasets:
                     logger.warning(f'Dataset {dataset_name} is not found')
-                    raise CidCritical(f'Dataset "{dataset_name}" ({ds.id}) is missing required fields. {(unmatched)}')
+                    raise CidCritical(f'Failed to find a Dataset "{dataset_name}" with required fields. Please check existing DataSets. '
```
yprikhodko, line 388: Add a check if it's executed in CFN Lambda or in console and use different error messages for both cases

cid/common.py
```diff
@@ -371,7 +371,7 @@ def _deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs
                         dataset_fields = {col.get('Name'): col.get('Type') for col in ds.columns}
                         required_fileds = {col.get('Name'): col.get('DataType') for col in source_template.datasets.get(dataset_name)}
                         unmatched = {}
-                        for k,v in required_fileds.items():
+                        for k, v in required_fileds.items():
                             if k not in dataset_fields or dataset_fields[k] != v:
                                 unmatched.update({k: {'expected': v, 'found': dataset_fields.get(k)}})
                         logger.debug(f'unmatched_fields={unmatched}')
```
yprikhodko, line 377: line 380 should be put inside else

cid/export.py
iakov-aws: done
image

cid/common.py
iakov-aws: done

yprikhodko (approved):

PR #480, opened 2023-02-07
Terraform modules for dashboard deployment and CUR setup

sean-nixon:
*Issue #, if available:* None

*Description of changes*:

1. Adds support for deploying CID dashboards via Terraform. This is implemented as a "wrapper" module that uses the `aws_cloudformation_stack` resource to deploy a CloudFormation stack from the `cid-cfn.yml` template. The template is too big to use an inline template, so instead it is uploaded to an S3 bucket which must be created in advance. Due to a limitation in the Terraform resource, Terraform can't detect changes to the template when using an S3 URL, so a workaround is implemented to append a "dummy" query parameter with the etag value of the S3 object. This will trigger the stack to update. Note, that Terraform will not show what will be updating in the CloudFormation Change Set, so Terraform output using this approach is limited.
2. Adds support for CUR setup using Terraform. There is a module for configuring the source account(s) and a module for configuring the destination accounts. This division allows for greater flexibility in deployment. The modules are native Terraform (no CloudFormation)

I used the tool [terraform-docs](https://terraform-docs.io/) to auto-generate documentation on the inputs, outputs, etc. and included the configuration file that I used for future updates.

---


terraform-modules/cid-dashboards/README.md
```diff
@@ -0,0 +1,151 @@
+# CID Terraform Module: cid-dashboards
+
+Terraform module to deploy CID dashboards. This module is a wrapper around CloudFormation
+to allow you to deploy CID dashboards using your existing Terraform workflows. Under the
+hood, the module will deploy a CloudFormation stack which will provision the necessary
+resources and a custom Lambda function to create the dashboards using `cid-cmd`.
+
+## Before You Start
```
iakov-aws, line 8: Can we point to official location? 

https://aws-managed-cost-intelligence-dashboards.s3.amazonaws.com/cfn/cur-aggregation.yaml

https://aws-managed-cost-intelligence-dashboards.s3.amazonaws.com/cfn/cid-cfn.yaml

terraform-modules/cid-dashboards/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+  - Existing S3 bucket to upload the CloudFormation template
```
iakov-aws, line 10: Not clear what key to use? Can we provide this in a readme?

sean-nixon: Can you clarify which key you are referring to?

iakov-aws (approved):

iakov-aws (approved):

PR #487, opened 2023-02-21
Fixed inconsistency when creating account_map view

dnabic-aws:
*Description of changes:*

* Changes related to creation of account_map from cid-cmd
  * SQL query for creation of view account_map as organization is moved from code into file cid/builtin/core/data/queries/shared/account_map_org.sql
  * Column names for view account_map are the same when this view is created as dummy and as organization
  * account_email_id column name is removed as it is not used anywhere


cid/builtin/core/data/queries/shared/account_map_dummy.sql
```diff
@@ -1,10 +1,14 @@
...
+CREATE OR REPLACE VIEW ${athena_view_name} AS
 SELECT DISTINCT
     line_item_usage_account_id                                     account_id,
-    MAX_BY(bill_payer_account_id,      line_item_usage_start_date) parent_account_id,
     MAX_BY(line_item_usage_account_id, line_item_usage_start_date) account_name,
-    MAX_BY(line_item_usage_account_id, line_item_usage_start_date) account_email_id
+    MAX_BY(bill_payer_account_id,      line_item_usage_start_date) parent_account_id,
+    CAST('' AS varchar)                                            account_status,
+    CAST('' AS varchar)                                            account_email,
+    CAST('' AS varchar)                                            account_email_id
```
iakov-aws, line 8: why we need both email_id and email?

cid/helpers/account_map.py
```diff
@@ -264,20 +263,17 @@ def create_account_mapping_sql(self, name) -> str:
...
+            resource_name='data/queries/shared/account_map_org.sql',
+        ).decode('utf-8'))
+
         accounts_sql = list()
         for account in self.accounts:
             acc = account.copy()
             account_name = acc.pop('account_name').replace("'", "''")
             accounts_sql.append(
-                """ROW ('{account_id}', '{account_name}:{account_id}', '{parent_account_id}', '{account_status}', '{account_email}')""".format(account_name=account_name, **acc))
+                """ROW ('{account_id}', '{account_name}:{account_id}', '{parent_account_id}', '{account_status}', '{account_email}', '{account_email}')""".format(account_name=account_name, **acc))
```
iakov-aws, line 276: {account_email} is there 2 times

dnabic-aws: Yes, second time is because it is used for account_email_id. But it might be better that account_email_id is ampty. Or if we agree to remove account_email_id column then this can be removed as well.

cid/builtin/core/data/queries/shared/account_map_dummy.sql
dnabic-aws: I am just trying to be very careful here. Email_id is not really used but I don't know if it was used somewhere in the past and if therefore we need some backwards compatibility.
But if we agree that this is not a valid concern, I would be happy to get rid of it.

dnabic-aws: email_id is removed

cid/helpers/account_map.py
dnabic-aws: second account_email is removed

cid/builtin/core/data/queries/shared/account_map.sql
```diff
@@ -1,10 +1,12 @@
 CREATE OR REPLACE VIEW account_map AS
 SELECT 
     ${account_id} AS account_id,
-    MAX_BY(${account_name}, concat(${account_name}, ': ', ${account_id}) AS account_name
-FROM 
+    MAX_BY(${account_name}, concat(${account_name}, ': ', ${account_id})) AS account_name,
```
dnabic-aws, line 4: Here is fixed missing closing bracket. Without fix, dashboard deployment will crash.

jannyg:
This PR seems stale? I ran into this issue when using the `cid-cmd map --account-map-source` to create mappings from CSV. Had to update views and datasets because of errors.

iakov-aws:
> This PR seems stale? I ran into this issue when using the `cid-cmd map --account-map-source` to create mappings from CSV. Had to update views and datasets because of errors.

Hi Please can you provide more info regarding your issue? This PR was not compatible with update procedure so we put it on hold.

jannyg:
> > This PR seems stale? I ran into this issue when using the `cid-cmd map --account-map-source` to create mappings from CSV. Had to update views and datasets because of errors.
> 
> Hi Please can you provide more info regarding your issue? This PR was not compatible with update procedure so we put it on hold.

I'm new to CUDOS and QuickSight so I'll try to explain as good as I can :)

After running the command (installed latest with pip) on CUDOS v0.2.46 the datasets in QuickSight failed with column errors (account_email_id not found) or STALE (mismatch between number of columns).

```
[Simba][AthenaJDBC](100071) An error has been thrown from the AWS Athena client. COLUMN_NOT_FOUND: line 12:227: Column 'account_email_id' cannot be resolved or requester is not authorized to access requested resources [Execution ID: 3e3864e5-59cf-44ec-a837-dff9c7a792fe]
```

I had to manually update all the views in Athena and refresh and publish the datasets in QuickSight to resolve the issue. I also have some NULL values in the dashboard. It seems like there could be some accounts maybe missing, so I will check that first. But I was expecting the behavior for mapping to show account number if no account_name was found.

iakov-aws:
[Cleanup close as inactive]

PR #523, opened 2023-04-19
used undeclared variable

dnabic-aws:
*Description of changes:*

* Variable `ds` is used before it is declared

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -355,7 +355,7 @@ class Cid():
             # First try to find the dataset with the id
             dataset = self.qs.describe_dataset(id=dataset_name)
             if isinstance(dataset, Dataset):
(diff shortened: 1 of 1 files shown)


iakov-aws (approved):

cid/common.py
```diff
@@ -355,7 +355,7 @@ def _deploy(self, dashboard_id: str=None, recursive=True, update=False, **kwargs
             # First try to find the dataset with the id
             dataset = self.qs.describe_dataset(id=dataset_name)
             if isinstance(dataset, Dataset):
-                logger.debug(f'Found dataset {dataset_name} with id match = {ds.arn}')
```
dnabic-aws, line 358: Variable ds doesn't exist before line 366

PR #553, opened 2023-06-19
Improving functionality of scheduling refresh for datasets

dnabic-aws:
* Added refresh schedule dependencies for datasets of kpi dashboard in resources.yaml
* Refresh schedule parameters are taken from config or from environment


cid/helpers/randtime.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_random_time_from_range(time_range):
+    """Return random time in format hh:mm within provided time range"""
+    if not time_range:
+        # return default random time between 02:00 and 04:59
+        return f"{random.randrange(2, 5):02d}:{random.randrange(0, 60):02d}", None
+
+    random_time = ""
+    try:
+        time_re = re.match(r'^(([01]\d|2[0-3]):([0-5]\d)|24:00)-(([01]\d|2[0-3]):([0-5]\d)|24:00)$', time_range)
```
yprikhodko, line 14: Could you please add comment explaining what is happening here and why we decided to do that

dnabic-aws: We wanted to avoid that all scheduled refreshes run at the same time. Therefore, in resources.yaml having something like this 

```
schedules:
  default:
    id: cid
...
    schedule_frequency_time_range: '02:00-05:00'
```

would mean that scheduled refresh time will be randomly selected between 2am and 5am. Therefore, the regex from above expects that `schedule_frequency_time_range` is provided in hh:mm-hh:mm format. 


However, writing this now, I am thinking that maybe we can also provide possibility for fixed time (instead of time range), so to allow also this:

```
schedules:
  default:
    id: cid
...
    schedule_frequency_time_range: '02:45'
```

What do you think?

iakov-aws: This looks like a fix?
`if time_to < time_from: time_to +=  timedelta(days=1)`

dnabic-aws: @iakov-aws  yes `if time_to < time_from: time_to +=  timedelta(days=1)` fixed it. Thanks!  Fix is pushed.

iakov-aws: Also i would try not to use regexp. You have 2 times, you can just split them and let str

```
timerange  = list (map(lambda x: x.strip(),  '23:09 - 02:22'.split('-')))
if len(timerange) == 1:
    return timerange[0]
elif len(timerange) == 2:
   ...
   # let strptime return a specific exception 
else:
   raise Exception('Input error ...')
```

dnabic-aws: Let's discuss this. Regex helps when provided input is hh:mm-hh:mm-hh:mm or anything like that. Also with split, I would have to check if there are only two elements in the list. So, with regex I don't need this extra logic.

cid/helpers/randtime.py
```diff
@@ -0,0 +1,25 @@
...
+        time_from = datetime.strptime(time_re.group(1), '%H:%M')
+        time_to = datetime.strptime(time_re.group(4), '%H:%M')
+        if time_to < time_from:
+            time_to += timedelta(days=1)
+        time_diff_sec = (time_to - time_from).total_seconds()
+        random_seconds = random.randrange(0, int(time_diff_sec))
+        random_time = (time_from + timedelta(seconds=random_seconds)).strftime('%H:%M')
+    except Exception as e:
```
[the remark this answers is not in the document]
iakov-aws: Yes but you can leave the raise here and catch it outside instead of if

iakov-aws:
@dnabic-aws i do not see an option to override timezone in parameters

cid/helpers/randtime.py
dnabic-aws: All recommendations are implemented

cid/helpers/randtime.py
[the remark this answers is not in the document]
dnabic-aws: all recommendations implemented.

cid/builtin/core/data/resources.yaml
```diff
@@ -123,26 +123,36 @@ datasets:
...
         - kpi_s3_storage_all
+      schedules:
+        - default
   kpi_tracker:
     File: kpi/kpi_tracker.json
     dependsOn:
       views:
         - kpi_tracker
+      schedules:
+        - default
```
iakov-aws, line 155: Should be schedule on the same level as dependsOn ?

dnabic-aws: That's actually a good point as it is not a dependency.

iakov-aws (approved):

PR #564, opened 2023-07-02
Feature scheduling strikes back

iakov-aws:
*Issue #, if available:*

*Description of changes:*


cid/builtin/core/data/resources.yaml
```diff
@@ -119,33 +127,45 @@ datasets:
       cur: true
       views:
         - ${cur_table_name}
+    schedules:
```
yprikhodko, line 130: Schedule on customer_all?

iakov-aws: Good point.

PR #617, opened 2023-09-07
Fix option_details column

selmanmoon:
*Issue #, if available:*
CONCAT function with null values are not working as it returns null if any of columns used in CONCAT function nulls, therefore added coalesce to keep null columns as 'na' and CONCAT function will still work as expected.

*Description of changes:*

COALESCE function added to all option_details columns.


iakov-aws:
Great work! Checking

iakov-aws (approved):

PR #656, opened 2023-10-28
allways allow access to data collection bucket for account map access

iakov-aws:
*Issue #, if available:*

*Description of changes:*


iakov-aws:
@yprikhodko @esc1144  please check an approve

esc1144 (approved):
Agree that these should be open out of the box. Having the new DS role at all is a very big improvement and our expansion of cross-over use cases will probably only increase over time.

PR #680, opened 2023-11-11
Add product storage fields and better messages if CUR does not contain fields

iakov-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/helpers/cur.py
+++ b/cid/helpers/cur.py
@@ -24,7 +24,7 @@ class CUR(CidBase):
         'line_item_line_item_type',
         'line_item_operation',
         'line_item_product_code',
-       #'line_item_resource_id',
+        'line_item_resource_id',
         'line_item_unblended_cost',
         'line_item_usage_account_id',
         'line_item_usage_amount',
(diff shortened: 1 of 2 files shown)


yprikhodko (approved):

cid/helpers/cur.py
```diff
@@ -11,21 +11,42 @@
...
+        'identity_line_item_id',
+        'identity_time_interval',
+        'line_item_legal_entity',
+        'line_item_line_item_description',
         'line_item_line_item_type',
-        'line_item_usage_start_date',
-        'line_item_usage_end_date',
+        'line_item_operation',
         'line_item_product_code',
+       #'line_item_resource_id',
```
yprikhodko, line 27: Why it's commented?

iakov-aws: because there was before a separate test on resource id when it was not needed in cudos datasets

iakov-aws: i put it back but we need to figure out a way to allow customer to install without it

PR #734, opened 2024-02-05
Changes to allow flexible views data, spriFile, spFile, and riFile in…

fsrich:
… cid/common.py

*Description of changes:* Change to allow flexibility while using resources.yaml to deploy external dashboards. If file contains view description with data, spriFile, spFile, and riFile. It will deploy views accordingly. 


iakov-aws:
as discussed  we will take another route

PR #765, opened 2024-03-28
rename rds extended support to extended support

yprikhodko:
*Issue #, if available:*

*Description of changes:*


iakov-aws (approved):

dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
```diff
@@ -1,10 +1,10 @@
 dashboards:
-  RDS EXTENDED SUPPORT COST PROJECTION V1.0.0:
```
iakov-aws, line 2: I would suggest not adding version as there can be side effects

PR #779, opened 2024-04-08
Update last_kpi_tracker_view.sql

sumitagarwalHcst:
Added PayerID joining condition in line 89,155,169,181 & 193

This is to fix duplicate linked account data in the scenario where a linked account is moved from one org to other.

*Issue #, if available:* Duplicate Linked account data

This is to cover the scenario where a linked account moves to another payer and we also have its historical cur data linked with its previous payer account, in that case we will have duplicate linked account as a result of our query which is causing this issue. To avoid that we can also include payerid with accountid so that we get distinct results for each payer and its linked account.

*Description of changes:*

Added PayerID in joining conditions to fix it.

### DIFF
--- a/cid/builtin/core/data/queries/kpi/last_kpi_tracker_view.sql
+++ b/cid/builtin/core/data/queries/kpi/last_kpi_tracker_view.sql
@@ -190,5 +190,5 @@ LEFT JOIN (
    FROM
      kpi_s3_storage_all
    GROUP BY 1, 2, 3
-)  s3_all ON ((s3_all.linked_account_id = account_id) AND (s3_all.billing_period = spend_all.billing_period) AND (s3_all.payer_account_id=spend_all.payer_account_id)))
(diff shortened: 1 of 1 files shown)


cid/builtin/core/data/queries/kpi/last_kpi_tracker_view.sql
```diff
@@ -191,5 +190,5 @@ LEFT JOIN (
    FROM
      kpi_s3_storage_all
    GROUP BY 1, 2, 3
-)  s3_all ON ((s3_all.linked_account_id = account_id) AND (s3_all.billing_period = spend_all.billing_period)))
+)  s3_all ON ((s3_all.linked_account_id = account_id) AND (s3_all.billing_period = spend_all.billing_period) AND (s3_all.payer_account_id=spend_all.payer_account_id)))
```
iakov-aws, line 193:
```suggestion
)  s3_all ON ((s3_all.linked_account_id = account_id) AND (s3_all.billing_period = spend_all.billing_period) AND (s3_all.payer_account_id = spend_all.payer_account_id)))
```

iakov-aws (approved):

PR #787, opened 2024-04-17
Aggregation CFN template for Data Exports

AWS-ZachErdman:
This commit adds a new cloudformation template to support CUR 2.0 in Data Exports with the same functionality as `cur-aggregation.yaml`


cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+              - s3:GetBucketAcl
+              - s3:GetBucketPolicy
+            Resource:
+              - !Sub 'arn:${AWS::Partition}:s3:::${SourceS3}'
+              - !Sub 'arn:${AWS::Partition}:s3:::${SourceS3}/*'
+          - Sid: AllowWriteBilling
+            Effect: Allow
+            Principal:
+              Service: 
+                - billingreports.amazonaws.com
```
yprikhodko, line 333: is this still needed for CUR2?

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+              - !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${ResourcePrefix}-CID-CUR2Creator"
+              - !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${ResourcePrefix}-CID-CUR2Creator:*"
+              - !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/${ResourcePrefix}-CID-CUR2Creator:*:*"
+        - PolicyName: "ExecutionSpecific"
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+            - Effect: Allow
+              Action:
+              - cur:PutReportDefinition
```
yprikhodko, line 516: do we need it for CUR2?

yprikhodko (commented):
Great work! I've added few comments inline. Also I see that PermissionsBoundary parameter and policy property excluded compare to current version. Is this intentional?  https://github.com/aws-samples/aws-cudos-framework-deployment/blob/02aec4cdda9e2e86e2b67b0e0ae315384a49be2a/cfn-templates/cid-cfn.yml#L211

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
...vation_normalized_units_per_reservation, reservation_number_of_reservations, reservation_recurring_fee_for_usage, reservation_reservation_a_r_n, reservation_start_time, reservation_subscription_id, reservation_total_reserved_normalized_units, reservation_total_reserved_units, reservation_units_per_reservation, reservation_unused_amortized_upfront_fee_for_billing_period, reservation_unused_normalized_unit_quantity, reservation_unused_quantity, reservation_unused_recurring_fee, reservation_upfront_value, resource_tags, savings_plan_amortized_upfront_commitment_for_billing_period, savings_plan_end_time, savings_plan_instance_type_family, savings_plan_net_amortized_upfront_commitment_for_billing_period, savings_plan_net_recurring_commitment_for_billing_period, savings_plan_net_savings_plan_effective_cost, savings_plan_offering_type, savings_plan_payment_option, savings_plan_purchase_term, savings_plan_recurring_commitment_for_billing_period, savings_plan_region, savings_plan_savings_plan_a_r_n, savings_plan_savings_plan_effective_cost, savings_plan_savings_plan_rate, savings_plan_start_time, savings_plan_total_commitment_to_date, savings_plan_used_commitment FROM COST_AND_USAGE_REPORT"
```
iakov-aws, line 76: Can we use YAML multiline? ex:
```
    CUR2:
      DefaultQuery: |
         SELECT 
            bill_bill_type, bill_billing_entity, bill_billing_period_end_date, bill_invoice_id, bill_invoicing_entity, bill_payer_account_id,
    ...
```

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+              Format: "PARQUET"
+              Compression: "PARQUET"
+              OutputType: "CUSTOM"
+        Name: !Ref ResourcePrefix
+        RefreshCadence:
+          Frequency: "SYNCHRONOUS"
+
+  # Deploy CUR2 via lambda due to missing cfn resource definition
+  # AWS::BCMDataExports::Export outside us-east-1
+  CUR2inUSEAST1:
```
iakov-aws, line 447: Probably need to change this from CUR2inUSEAST1 to CUR2viaCustomResource or something similar as there is also cn-* region.

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+            reason: "One Time execution. No need for ReservedConcurrentExecutions"
+
+  CIDLambdaAnalyticsExecutor:
+    Type: Custom::CIDLambdaAnalyticsExecutor
+    Properties:
+      ServiceToken: !GetAtt CIDLambdaAnalytics.Arn
+      DeploymentType:
+        - Fn::If:
+          - IsDestinationAccount
+          - "cid-cur-aggregator"
```
iakov-aws, line 737: cid-dataexport-aggregator

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+          - "cid-cur-replication"
```
iakov-aws, line 738: cid-dataexport-replication

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,749 @@
...
+      ReplicationConfiguration:
+        Role: !GetAtt CrossRegionReplicationRole.Arn
+        Rules:
+          -
+            Destination:
+              Bucket:
+                Fn::Sub: "arn:${AWS::Partition}:s3:::${ResourcePrefix}-${DestinationAccountId}-shared"
+              StorageClass: STANDARD
+            Id: ReplicationRule1
+            Prefix: ""
```
iakov-aws, line 268: Need to manage this carefully. Please can you test creating athena table with 2 aggregated exports?

yprikhodko: +1 it should be leading to data folder to exclude metadata folder replication

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: `billingreports.amazonaws.com` is not needed for Data Exports/CUR2 but we've been adding it in our UI workflow in case customers are trying to deliver Legacy CURs to the same bucket.

We could remove it here if we think the bucket would only be used for this.

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: Yes, `cur:PutReportDefinition` is required for creating a CUR 2.0 in Data Exports. This was implemented to prevent CUR 2 from becoming a backdoor way to access CUR data. For example, if a management account had an SCP blocking `cur:PutReportDefinition` and then we launched CUR 2, they could get access to those APIs and create a CUR.

AWS-ZachErdman:
> Great work! I've added few comments inline. Also I see that PermissionsBoundary parameter and policy property excluded compare to current version. Is this intentional?
> 
> https://github.com/aws-samples/aws-cudos-framework-deployment/blob/02aec4cdda9e2e86e2b67b0e0ae315384a49be2a/cfn-templates/cid-cfn.yml#L211

@yprikhodko I had it in the version you shared with me, but it wasn't in the version that was present on Well Architected Labs website. It also kept causing errors for me when I tried to upload the template if I didn't specify anything in that field so I ended up just removing it.

We can add it back in if you'd like and know how to make it not error.

Screenshot 2024-04-14 at 7 37 25 PM

cfn-templates/data-exports-aggregation.yaml
yprikhodko: I think mixing CUR legacy and CUR 2 in the same bucket can cause a lot of issues with Athena and Glue. I don't think it worth to consider this use case for CID template

AWS-ZachErdman: Ah makes sense - okay will remove

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: makes sense - will fix

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: fixed

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: fixed

cfn-templates/data-exports-aggregation.yaml
AWS-ZachErdman: Should it be a filter on the ReplicationConfiguration? Or how else would I exclude metadata folder replication?

iakov-aws:
This might happen if you delete s3 in one region and then try to recreate in another one too soon with the same name. You can try changing prefix or wait a bit more.

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,743 @@
...
+              AWS:
+                Fn::If:
+                  - EmptySourceAccountIds
+                  - !Ref AWS::AccountId
+                  - !Split [',', !Ref SourceAccountIds]
+            Action:
+              - s3:ReplicateDelete
+              - s3:ReplicateObject
+            Resource:
+              - Fn::Sub: 'arn:${AWS::Partition}:s3:::${DestinationS3}/*'
```
iakov-aws, line 199: probably need to add here a permission only to copy `*/data/*`

cfn-templates/data-exports-aggregation.yaml
iakov-aws: you can try both. whatever works the best. I would suggest using the right prefix would be optimal as you know the path. Also this means that we need to use replication even if we have a local CUR (or alternatively prohibit cur to write */metadata/*).

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,743 @@
...
+        FROM COST_AND_USAGE_REPORT
```
iakov-aws, line 78: Need to integrate here SCAD columns?

petrokashlikov: Integrated as parameter

cfn-templates/data-exports-aggregation.yaml
petrokashlikov: As discussed, we will have to use local replication, because: 1) S3 replication itself doesn't support exclusion 2) We can't exclude metadata folder in Athena 3) SCAD won't accept reduced bucket write permissions without metadata folder

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,1091 @@
...
+                      Export=export
+                  )
+                  print(json.dumps(res))
+              elif event['RequestType'] == 'Update':
+                  old_export_name = event['OldResourceProperties']['Export']['Name']
+                  if export["Name"] != old_export_name:
+                      res = client.create_export(
+                          Export=export
+                      )
+                      print(json.dumps(res))
```
iakov-aws, line 746: do you need to delete old one as well?

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,1091 @@
...
+        rules_to_suppress:
+          - id: 'W89'
+            reason: "This Lambda does not require VPC"
+          - id: 'W92'
+            reason: "One Time execution. No need for ReservedConcurrentExecutions"
+
+###########################################################################
+# Glue Database for CID framework
+###########################################################################
+  CIDDatabase:
```
iakov-aws, line 786: This can be a blocker. I struggle to understand the full flow if CID database is defined here

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,1091 @@
...
+          Location: !Sub "s3://${DestinationS3}/cur2/"
+          NumberOfBuckets: -1
+          InputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat
+          OutputFormat: org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat
+          SerdeInfo:
+            Parameters:
+              serialization.format: '1'
+            SerializationLibrary: org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe
+          StoredAsSubDirectories: false
+          Columns: # All fields required for CID
```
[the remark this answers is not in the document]
iakov-aws: @petrokashlikov can you check this one?

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,1091 @@
...
+            - {"Name": "line_item_usage_end_date", "Type": "timestamp" }
+            - {"Name": "line_item_usage_start_date", "Type": "timestamp" }
+            - {"Name": "line_item_usage_type", "Type": "string" }
+            - {"Name": "pricing_lease_contract_length", "Type": "string" }
+            - {"Name": "pricing_offering_class", "Type": "string" }
+            - {"Name": "pricing_public_on_demand_cost", "Type": "double" }
+            - {"Name": "pricing_purchase_option", "Type": "string" }
+            - {"Name": "pricing_term", "Type": "string" }
+            - {"Name": "pricing_unit", "Type": "string" }
+            - {"Name": "product", "Type": "map" }
```
yprikhodko, line 855:
```suggestion
            - {"Name": "product", "Type": "map<string,string>" }
```

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,1091 @@
...
+            - {"Name": "savings_plan_amortized_upfront_commitment_for_billing_period", "Type": "double" }
+            - {"Name": "savings_plan_end_time", "Type": "timestamp" }
+            - {"Name": "savings_plan_offering_type", "Type": "string" }
+            - {"Name": "savings_plan_payment_option", "Type": "string" }
+            - {"Name": "savings_plan_purchase_term", "Type": "string" }
+            - {"Name": "savings_plan_savings_plan_a_r_n", "Type": "string" }
+            - {"Name": "savings_plan_savings_plan_effective_cost", "Type": "double" }
+            - {"Name": "savings_plan_start_time", "Type": "timestamp" }
+            - {"Name": "savings_plan_total_commitment_to_date", "Type": "double" }
+            - {"Name": "savings_plan_used_commitment", "Type": "double" }
```
yprikhodko, line 877:
```suggestion
            - {"Name": "savings_plan_used_commitment", "Type": "double" }
            - {"Name": "cost_category", "Type": "map<string,string>" }
            - {"Name": "discount", "Type": "map<string,string>" }
            - {"Name": "resource_tags", "Type": "map<string,string>" }
```

cfn-templates/data-exports-aggregation.yaml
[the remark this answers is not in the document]
petrokashlikov: @iakov-aws Checking

cfn-templates/data-exports-aggregation.yaml
petrokashlikov: Added

cfn-templates/data-exports-aggregation.yaml
petrokashlikov: Added

cfn-templates/data-exports-aggregation.yaml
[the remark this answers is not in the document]
petrokashlikov: Done

cfn-templates/data-exports-aggregation.yaml
iakov-aws: in case when you change the name on update and you have to create the new CUR, you need to delete old. Correct ?

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -0,0 +1,901 @@
...
+      ServiceToken: !GetAtt CidLambdaAnalytics.Arn
+      DeploymentType:
+        - Fn::If:
+          - IsDestinationAccount
+          - "cid-dataexport-aggregator"
+          - "cid-dataexport-replication"
+
+Outputs:
+  GovernanceBucketName:
+    Description: Bucket where to store and aggregate CUR2
```
iakov-aws, line 894:
```suggestion
    Description: Bucket with aggregate Data Exports
```

iakov-aws (approved):

iakov-aws:
Great job @AWS-ZachErdman  and @petrokashlikov ! Awesome contribution that will help many customers.

PR #807, opened 2024-04-30
CUR proxy + CUR 2.0 - Major v4.0.0

iakov-aws:
*Issue #, if available:*

*Description of changes:*


cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
...
+    'product_finding_source': "product['finding_source']",
+    'product_finding_storage': "product['finding_storage']",
+    'product_standard_group': "product['standard_group']",
+    'product_standard_storage': "product['standard_storage']",
+    'product_product_type': "product['product_type']",
+    'product_pricingplan': "product['pricingplan']",
+    'product_provider': "product['provider']",
+    'product_subservice': "product['subservice']",
+    'product_type': "product['type']",
+    'product_tickettype': 'n/a',
```
petrokashlikov, line 170: Why this fields are mapped to n/a?  What makes you think they don't exist in CUR 2.0?

cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
...
+                    return 'cast(NULL AS MAP<VARCHAR, VARCHAR>)'
+                return f'''
+                    MAP(
+                        ARRAY[{', '.join(["'" + key + "'" for key in map_mapping.keys()])}],
+                        ARRAY[{', '.join([cur1_field for cur1_field in map_mapping.values()])}]
+                    )
+                '''
+            cur1to2mapping = {value: key for key, value in cur2to1mapping.items()}
+            if field in cur1to2mapping:
+                return f'{cur1to2mapping.ge(field, field)}'
```
petrokashlikov, line 382: probably typo, should be .get

petrokashlikov (commented):
Submitted few comments. Also wasn't able to understand full logic as of now.

cid/helpers/cur_proxy.py
iakov-aws: In my account I have cur1 and cur2 so I pulled all fields of both and those were missing.

cid/helpers/cur_proxy.py
iakov-aws: Yes

cid/helpers/cur_proxy.py
petrokashlikov: I think it is because you for example don't have resources with extended support for this column for example product_extended_support_pricing_year.  In reality I think they are all mapped 1:1 to old fields.

cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
...
+    'product_with_active_users': 'n/a',
+    'product_abd_instance_class': 'n/a',
+    'product_size_flex': 'n/a',
+    'product_engine_major_version': 'n/a',
+    'product_extended_support_pricing_year': 'n/a',
+    "concat('name', bill_payer_account_id)": 'bill_payer_account_name',
+    "concat('name', line_item_usage_account_id)": 'line_item_usage_account_name',
+}
+
+default_columns = {
```
petrokashlikov, line 182: What is purpose of default columns compared to mapping above?

cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
...
+
+    logging.basicConfig(level=logging.INFO)
+    logger.setLevel(logging.DEBUG)
+    athena = Athena(session=boto3.Session())
+    glue = Glue(session=boto3.Session())
+    cur = CUR(athena=athena, glue=glue)
+    proxy = ProxyView(
+        cur=cur,
+        target_cur_version='2',
+        fields_to_expose = {
```
petrokashlikov, line 441: What purpose does this serve? And why we hard code this here, but at line 341 we are joining this with default columns? Can't we just add them to default_columns? 

self.fields_to_expose = list(set(default_columns[self.target_cur_version] + (fields_to_expose or []) ))

cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
...
+
+    def create_or_update_view(self):
+        self.read_from_athena()
+        all_fields  = list(set(self.exposed_fields + self.fields_to_expose))
+        lines = []
+        print('all_fields',  all_fields)
+        for field in all_fields:
+            field_type = self.cur.get_type_of_column(field)
+            mapped_expression = self.get_sql_expression(field, field_type)
+            requirement = mapped_expression.split('[')[0]
```
petrokashlikov, line 401: What is purpose of this?

cid/helpers/cur_proxy.py
```diff
@@ -0,0 +1,452 @@
+import re
+import logging
+
+logger = logging.getLogger(__name__)
+
+cur2to1mapping = {
```
petrokashlikov, line 6: Are we supposed to create 3 more of this for each scenario?
CUR1 -> CUR1_proxy 
CUR1 -> CUR2_proxy 
CUR2 -> CUR2_proxy

petrokashlikov (commented):
Added more questions/comments

cid/helpers/cur_proxy.py
iakov-aws: Mapping above contains all 
Default - the most used.

cid/helpers/cur_proxy.py
iakov-aws: This is a test code.

cid/helpers/cur_proxy.py
iakov-aws: For a field product['something'] the requirement would be product

cid/helpers/cur_proxy.py
iakov-aws: Trying to avoid this and reuse to maximum this mapping

cid/helpers/cur_proxy.py
petrokashlikov: And why we need this from logic perspective? Any specific benefits it brings?

iakov-aws: With a reasonable set of default dashboards authors do not need to care about describing cur fields. 

cid-cmd export cannot do that today, so it stays a manual operation for authors.

petrokashlikov (approved):
LGTM

yprikhodko (approved):

PR #812, opened 2024-05-09
Filter scad from views

rarmist:
*Issue #, if available:*

*Description of changes:*
Split Allocation Cost Data (SCAD) is not necessary for summary_view, hourly_view, and resource_view as the cost is a subset of other usage. SCAD also uses new fields to visualize cost which is only used in the new SCAD dashboard. 


yprikhodko (commented):
Could you please add the same to the rest of summary_view* hourly_view and resource_view* files?

iakov-aws:
@rarmist  Please check all the files where summary view, hourly view and resource view are defined (there are typically 4 optons)

rarmist:
I updated all the views to include the SCAD filter.

iakov-aws (approved):

PR #830, opened 2024-06-07
Implement new AWS Feeds Dashboard in Amazon QuickSight

ftresgallo:
Implement new AWS Feeds Dashboard in Amazon QuickSight


dashboards/aws-feeds/aws-feeds.yaml
```diff
@@ -0,0 +1,2828 @@
...
+    dependsOn:
+      views:
+      - aws_feeds
+    schedules:
+    - default
+crawlers: {}
+views:
+  aws_feeds:
+    data: |-
+      CREATE OR REPLACE VIEW ${athena_database_name}.aws_feeds AS
```
yprikhodko, line 2791: Any reason why not to create it in ${data_collection_database_name} ?

ftresgallo: You are right, I have removed the param as being duplicated.

iakov-aws: The reason is to align with other dashboards like extended support and others that use cur and data collection 

I would recommend keeping 2 databases. They will be autodetected soon.

yprikhodko: I don't see any benefit of having them align, but if you think this will help I'm fine

dashboards/aws-feeds/aws-feeds.yaml
```diff
@@ -0,0 +1,2823 @@
...
+    dependsOn:
+      datasets:
+      - aws_feeds
+      - aws_feeds_blog_post
+      - aws_feeds_security_bulletin
+      - aws_feeds_whats_new
+      - aws_feeds_youtube
+    name: AWS Feeds
+    dashboardId: aws-feeds
+    category: Custom
```
iakov-aws, line 12:
```suggestion
    category: Advanced
```

iakov-aws (approved):

PR #825, opened 2024-06-01
add conditions for cur s3 policies

iakov-aws:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #879, opened 2024-07-08
Attempt fixing S3 resource policy introduced in #825

eriweb:
PR #825 introduced new iam conditions, one of them uses wildcard without Like operator

To be honest I don't even know if the SourceArn is required/valid, but we're currently troubleshooting missing cur reports and making the policy valid is first step.

Right now cur reports issues with S3 bucket permissions

*Issue #, if available:*

*Description of changes:*


iakov-aws (approved):

iakov-aws:
Thank you @eriweb !

PR #887, opened 2024-07-10
Version detection for definition based dashboards

juchavw:
*Issue #, if available:*

*Description of changes:*
Adding logic and components to support version detection for definition based dashboards. This will help support the status cid-cmd sub-command and associated update and refresh actions. This PR includes:
- A new module with the specification for a dashboard definition to hold details in-memory during the execution of the CLI tool.
- Adjustments to the dashboard discovery process to load dashboard definition details, resolve current and latest deployed versions and establish dashboard health and status based on the currently deployed version and the latest version available.
- Adjusting refresh mechanism to ensure dashboard definition versions are changed right after a dashboard update and display the latest version is now the currently deployed version.
- Minor adjustments to the status sub-command output
  - Removing VersionId value from the status output. The VersionId is still used for TEMPLATE based dashboards but it does not need to be displayed for the CLI users.


cid/helpers/quicksight/version.py
```diff
@@ -0,0 +1,63 @@
...
+        return f'v{self.major}.{self.minor}.{self.build}'
+    
+    def _parse(self, str_version):
+        
+        version_pattern = re.compile(r"^[v|V](?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.(?P<build>[0-9]+)$")
+        results = version_pattern.match(str_version)
+        
+        if not results:
+            logger.debug(f'Could not find version pattern in provided string: {str_version}')
+            major = 1
```
juchavw, line 27: This is an interim workaround to cater for all template based dashboards that may not follow the right convention for the dashboard version ( v<major>.<minor>.<build> ) and avoid them from breaking the cid-cmd execution.

cid/helpers/quicksight/definition.py
```diff
@@ -0,0 +1,68 @@
...
+            if version_matches:
+                return version_matches[0]
+            else:
+                version_matches = re.findall(r"(v\d+?\.\d+?)", all_about_content)
+                if version_matches:
+                    return f"{version_matches[0]}.0"
+
+        return None
+    
+    def _extract_sheet_visuals_content(self, sheets: list):
```
juchavw, line 51: The health dashboard uses insights visuals instead of text boxes. This can be revisited to review how to perform version resolution for deployed dashboards without relying on their definitions or template description or similar information.

cid/helpers/quicksight/definition.py
```diff
@@ -0,0 +1,68 @@
...
+        insight_visuals_content = [
+            visual["InsightVisual"]["InsightConfiguration"]["CustomNarrative"].get("Narrative", "")
+            for visual in visuals
+            if "InsightVisual" in visual
+            and "InsightConfiguration" in visual["InsightVisual"]
+            and "CustomNarrative" in visual["InsightVisual"]["InsightConfiguration"]
+        ]
+        return insight_visuals_content
+
+    def _extract_sheet_textboxes_content(self, sheets: list):
```
juchavw, line 63: Most definition based dashboards are using text boxes for the "About" sheet content, where the version string is located.

cid/helpers/quicksight/__init__.py
```diff
@@ -256,9 +261,23 @@ def _safe_int(value, default=None):
                     logger.info(f'Unable to describe template for {dashboardId}, {exc}')
             else:
                 logger.info("Minimum template version could not be found for Dashboard {dashboardId}: {_template_arn}, deployed template could not be described")
+        
+        if _template_arn is None:
+            # Resolve source definition (the latest definition publicly available)
+            data_stream = io.StringIO(_definition["data"])
```
iakov-aws, line 267: are we sure that data is in the definition?

cid/helpers/quicksight/__init__.py
```diff
@@ -256,9 +261,23 @@ def _safe_int(value, default=None):
                     logger.info(f'Unable to describe template for {dashboardId}, {exc}')
             else:
                 logger.info("Minimum template version could not be found for Dashboard {dashboardId}: {_template_arn}, deployed template could not be described")
+        
+        if _template_arn is None:
```
iakov-aws, line 265: ``
```suggestion
        elif _template_arn is None and "data" in _definition:
```

iakov-aws: need also add `else`to throw a debug message that this dashboard is not template and not definition based

cid/helpers/quicksight/__init__.py
```diff
@@ -256,9 +261,23 @@ def _safe_int(value, default=None):
...
+            # Resolve source definition (the latest definition publicly available)
+            data_stream = io.StringIO(_definition["data"])
+            definition_data = yaml.safe_load(data_stream)
+            dashboard.sourceDefinition = CidQsDefinition(definition_data)
```
iakov-aws, line 269: 1/ _definition["data"] can be a dict or string
2/ both theoretically can be templates as per this code:

https://github.com/aws-samples/aws-cudos-framework-deployment/blob/d2865cb42e50ff836bd44d9ce635a789d3605536/cid/common.py#L482C1-L489C62

so ideally would be great to have it as one piece of code. I do not see how it can be done simply on the level of helper quicksight as parameters on the common.py.

cid/common.py
```diff
@@ -692,7 +692,11 @@ def status(self, dashboard_id, **kwargs):
                         logger.info(f'Updating dashboard: {dashboard.id} with Recursive = {recursive}')
                         self._deploy(dashboard_id, recursive=recursive, update=True)
                         logger.info('Rediscover dashboards after update')
-                        self.qs.discover_dashboards()
+                        
+                        refresh_overrides = {
```
iakov-aws, line 696: why not just set the list?

juchavw: you are right, a list would be enough, in the end the original one was a flag so, no need for a dictionary unless there are plans to have more functionality and context for a refresh action.

juchavw: refresh_overrides has been changed to a list now.

PR #904, opened 2024-07-21
fix data collection database management

iakov-aws:
*Issue #, if available:*

*Description of changes:*


petrokashlikov (approved):
LGTM

dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
```diff
@@ -3864,6 +3866,8 @@ views:
           description: "Enter the name of the CUR table"
           global: True
         data_collection_database_name:
+          type: athena
+          query: SELECT DISTINCT table_schema FROM information_schema.columns WHERE table_name = 'inventory_eks_data'
```
juchavw, line 3870: This one depends on 2 tables: inventory_rds_db_instances_data and inventory_rds_clusters_data inventory_rds_db_clusters_data.

Is the 'query' param only referring to 'inventory_eks_data' because it assumes that if it exists then the other ones exist?

Otherwise, perhaps changing the table name to be one of the 2 this view uses could be enough for now, unless there's a way to validate both of them (instances and clusters):

```
SELECT DISTINCT table_schema FROM information_schema.columns WHERE table_name = 'inventory_rds_db_instances_data'
```

dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
```diff
@@ -3867,6 +3869,8 @@ views:
...
```
iakov-aws, line 3873:
```suggestion
          query: SELECT DISTINCT table_schema FROM information_schema.columns WHERE table_name = 'inventory_rds_db_instances_data'
```

juchavw (approved):

Issue #989, opened 2024-10-29
CUR export bucket s3 policy


vivien-ramahandry-fox:
Since the this #825 and also this one #879 AWS service billing reports can't upload new CUR on my bucket

iakov-aws:
Thanks for reporting this. 

Does this work if you remove the condition from the statement?

vivien-ramahandry-fox:
Hello @iakov-aws I deleted yesterday and It works this morning
and before I have this
![image](https://github.com/user-attachments/assets/0951c73d-26db-4e24-abab-194fefb4298b)

iakov-aws:
Thanks 

Normally we follow documentation 

https://docs.aws.amazon.com/cur/latest/userguide/cur-s3.html


Was will check internally and get back to you.

iakov-aws:
Can you confirm you just deleted account restriction?

https://github.com/aws-samples/aws-cudos-framework-deployment/blob/main/cfn-templates/cur-aggregation.yaml

vivien-ramahandry-fox:
I deleted the 2 conditions here
https://github.com/aws-samples/aws-cudos-framework-deployment/blob/main/terraform-modules/cur-setup-destination/main.tf#L186
because I'm using terraform module for cur part and not cloudformation

iakov-aws:
Ok I see. We updated cloud formation but not TF. 

Do you want to suggest a PR just leaving account condition. If not we will do it in the following days. 

Thank again for raising and your analysis!

vramahandry:
Yeah sure I will create the PR with my personal account and close the other one but I think you can add the region us-east-1 on the condition. I will test it and let you know

vramahandry:
you can close this #990

iakov-aws:
Thanks for update and your contribution!

PR #991, opened 2024-10-30
fix(tf): cur bucket policy add us-east-1

vramahandry:
*Issue #989, if available:*

*Description of changes:*
Change the region to be us-east-1 for contions in cur bucket policy like the documentation [here](https://docs.aws.amazon.com/cur/latest/userguide/cur-s3.html) because cur definition and billing are just in us-east-1


iakov-aws (approved):

iakov-aws:
Looks great TY!

Issue #1029, opened 2024-11-15
Terraform deployment not compatible with CID dashboard deployment


nickcasscs:
When deploying either the CUDOS, Cost Intelligence or KPI dashboards the CID Dashboard [deployment via terraform](https://github.com/aws-samples/aws-cudos-framework-deployment/tree/main/terraform-modules#step-3-dashboard-setup) wrapper uses a cloudformation template that relies on a [stack output](https://github.com/aws-samples/aws-cudos-framework-deployment/blob/2748973eecec04fcac5c20bcb6ceac3485f9ba9b/cfn-templates/cid-cfn.yml#L1768) `cid-DataExport-Database` from the [cloudformation destination deployment](https://github.com/aws-samples/aws-cudos-framework-deployment/blob/2748973eecec04fcac5c20bcb6ceac3485f9ba9b/cfn-templates/data-exports-aggregation.yaml#L1457). When deploying the solution using the `cur-setup-destination` terraform module this stack doesn't exist and as a result, the stack output also does not exist; this causes the stack deploy to fail.

By passing `CurVersion = "1.0"` to the cid deployment stack parameters this issue looks like it is resolved. 
There isn't any reference in the documentation that this stack parameter is required currently in either the Terraform deployment instructions or the README for the cid-dashboards terraform module. 

It would save implementers a lot of time spent troubleshooting if this is noted in the documentation.

iakov-aws:
@sean-nixon we probably need to provide terrarform wrappers for Data Exports modules as well

christrotter:
Also running into something similar.  We are being advised by our TAM to go forward with CUDOS v5 and CUR 2.0, but the 'source' Terraform module provided only creates the `aws_cur_report_definition` resource: https://github.com/aws-samples/aws-cudos-framework-deployment/blob/main/terraform-modules/cur-setup-source/main.tf#L253

That resource is legacy and not CUR 2.0.

Rather, the resource it should be using is this:`aws_bcmdataexports_export` 
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/bcmdataexports_export

For us to move forward, we will have to copy the source module and modify it so that it uses the correct resource.  However - I think that once that issue is addressed, the rest of the provided bits and pieces should work, as all of the complicated downstream infrastructure has already been coded to support CUR 2.0.

christrotter:
Hopefully this is helpful.  I followed the contributions guide as best I could.

[...]

PR #1040, opened 2024-11-22
Add cur2 tf source module

christrotter:
*Issue #, if available:*
https://github.com/aws-samples/aws-cudos-framework-deployment/issues/1029

*Description of changes:*
For those who are starting their CID/CUDOS journey, the advice from AWS TAMs and documentation is that CUDOS v4 is deprecated and no longer supported.  Furthermore, CUR 1.0 is similarly deprecated and no longer supported.  Therefore, the recommended starting point is via CUDOS v5 and CUR 2.0.

The key point of this change is that the source account (payer account) - i.e. where you want to set up the Data Export - is currently only able to build a Legacy CUR 1.0 export (the `aws_cur_report_definition` resource).  In order to use CUR 2.0, you must use the new resource type `aws_bcmdataexports_export`.

This new resource type, which I have made the default, also requires the following changes:
- IAM policy changes to allow the new AWS service to access it
- an output to match the old resource
- logic variable so you can force CUR 1.0 if needed
- a variable to set a custom query for report fields (defaulting to all, as per AWS recommendation)
- readme info to help advise users of potential pitfalls (e.g. cost)

Also, I was unfortunately not able to run the `bats` tests locally due to account restrictions, but added some relevant changes that I think are what you'd want?

Finally, there is one item I still need to validate, and that is how the rest of the CFN stack handles the S3 pathing - I think it should be fine?


sean-nixon (changes requested):
Apologies for the delay reviewing this. I started adding some inline feedback; however, as you mentioned on the linked issue, I don't think retrofitting the existing modules to work with CURv2 will work. Instead, I think we will need to create new modules specifically for data exports (e.g. `data-exports-source` and `data-exports-destination` modules), effectively converting the CFN template [data-exports-aggregation.yaml](https://github.com/aws-samples/aws-cudos-framework-deployment/blob/main/cfn-templates/data-exports-aggregation.yaml).

If you have a working version and want to contribute, I definitely encourage you to submit another PR which we can use as a base for getting it this functionality merged into mainline.

Some things to consider:
* FOCUS support
* Cost Optimization Hub data support
* S3 bucket naming consistent with CFN

PR #1053, opened 2024-12-05
Updates the Lambda function that creates the service-linked role for …

jaehyi-aws:
## Description of changes:*
The `Timeout` for the Lambda function was 15 seconds, which would not have been enough for the function to complete execution because of 30-second sleep in the code.

1. Increases the timeout to 90 seconds.
2. Adds `ServiceTimeout` to the custom resource to match the function timeout.
3. Modularizes the code within the function for readability.
4. Uses ListEnrollmentStatuses, instead of GetPreferences, to check if an account is enrolled in COH.
5. Uses standard mode for retries, with the max attempts capped to 5.

### DIFF
--- a/cfn-templates/data-exports-aggregation.yaml
+++ b/cfn-templates/data-exports-aggregation.yaml
@@ -637,7 +637,7 @@ Resources:
                 Resource: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:role/aws-service-role/bcm-data-exports.amazonaws.com/AWSServiceRoleForBCMDataExports'
               - Effect: Allow
                 Action:
-                  - cost-optimization-hub:GetPreferences
+                  - cost-optimization-hub:ListEnrollmentStatuses
                 Resource: '*' # Cannot restrict this
 
     Metadata:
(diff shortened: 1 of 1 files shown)


cfn-templates/data-exports-aggregation.yaml
```diff
@@ -657,57 +657,96 @@ Resources:
...
-                      print("Additional wait to make sure the role is available")
-                      time.sleep(30)
-
-                  cfnresponse.send(event, context, cfnresponse.SUCCESS, {})
+                  account_id = context.invoked_function_arn.split(":")[4]
+                  validate_coh_enrollment(account_id)
+                  create_service_linked_role(
+                      service_name="bcm-data-exports.amazonaws.com",
+                      description="Service-linked role for bcm-data-exports.amazonaws.com"
+                  )
```
iakov-aws, line 697: Please keep the sleep as if we create export right away after creating it can fail.

iakov-aws (changes requested):

cfn-templates/data-exports-aggregation.yaml
```diff
@@ -657,57 +657,96 @@ Resources:
...
+              try:
+                  logger.info(f"Creating a service-linked role for {service_name}...")
+
+                  role_name = iam_client.create_service_linked_role(
+                      AWSServiceName=service_name,
+                      Description=description
+                  )["Role"]["RoleName"]
+                  waiter = iam_client.get_waiter("role_exists")
+                  waiter.wait(RoleName=role_name)
+                  time.sleep(10)  # Additional wait time, just in case
```
jaehyi-aws, line 740: This ensures that the role is available after creation: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam/waiter/RoleExists.html

cfn-templates/data-exports-aggregation.yaml
jaehyi-aws: There is `wait` and `sleep` inside the `create_service_linked_role` function: https://github.com/aws-samples/aws-cudos-framework-deployment/pull/1053/files#r1871433207

iakov-aws (approved):

iakov-aws (approved):

iakov-aws:
great contribution of high quality code! Thanks!

PR #1054, opened 2024-12-05
Makes parameters of str type case-insensitive

jaehyi-aws:
## Description of changes:

Updates to `get_yesno_parameter` function:
1. Adds type hints to function parameters
2. Attempts to simplify the logic within the function
3. Uses identity comparison ( `is` ) in place of equality ( `==` ) with `None`
4. Converts the `default` value (once it is a `str` type) to lowercase within the function, ensuring case-insensitive comparison regardless of how the `default` is specified.

### DIFF
--- a/cid/utils.py
+++ b/cid/utils.py
@@ -173,15 +173,18 @@ def get_parameters():
     return dict(params)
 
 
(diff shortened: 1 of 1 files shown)


iakov-aws (approved):

iakov-aws:
LGTM. Thanks

PR #1068, opened 2024-12-16
Fix billing_period splitting in kpi_ebs_snap_view sql query

cedricweider:
*Issue #, if available:*
Athena view kpi_ebs_snap doesn't give any output due to the bug we fixed here. 

*Description of changes:*
I replaced the splitting sign for billing_period in lines 11 and 12 of kpi_ebs_snap_view SQL query because the original code doesn't output any result from CUR 2.0 data export.
Originally the code tries to split the billing_period var using a forward slash (/) and we replaced it with a dash (-).


iakov-aws:
Great catch!

PR #1075, opened 2024-12-19
yaml file copies added into s3 prefix with CID_VERSION

vtaws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/assets/publish_lambda_layer.sh
+++ b/assets/publish_lambda_layer.sh
@@ -3,8 +3,8 @@
 
 # First build layer
 layer=$(./assets/build_lambda_layer.sh)
-# Source the version from the previous script execution
-source ./assets/build_lambda_layer.sh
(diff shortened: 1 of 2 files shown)


assets/publish_lambda_layer.sh
```diff
@@ -3,6 +3,9 @@
 
 # First build layer
 layer=$(./assets/build_lambda_layer.sh)
+# Source the version from the previous script execution
+source ./assets/build_lambda_layer.sh
```
iakov-aws, line 7: this is a heavy script i would just copy the line with version or get it from version.json

vtaws: Replaced source with exact line which generates the version.

iakov-aws (approved):

iakov-aws (approved):

PR #1077, opened 2024-12-20
fix null table error in cur detection

iakov-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/helpers/athena.py
+++ b/cid/helpers/athena.py
@@ -595,7 +595,7 @@ class Athena(CidBase):
         ''')
 
     def find_tables_with_columns(self, columns: list, database_name: str=None, catalog_name: str=None, max_items: int=10000):
-        """ This function searches a table with a given set of columns
+        """ This function return interator of tables with a given set of columns.
         """
         iterator = self.client.get_paginator('list_table_metadata').paginate(
             DatabaseName=database_name or self.DatabaseName,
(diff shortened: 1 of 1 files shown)


cid/helpers/athena.py
```diff
@@ -604,6 +604,9 @@ def find_tables_with_columns(self, columns: list, database_name: str=None, catal
                 'MaxItems': max_items, # sometimes customers can have 1'000s of tables (due to a crawler going crazy for example)
             },
         )
-        return iterator.search(f"""
-            TableMetadataList[?{' && '.join(["contains(Columns[].Name, '"+col+"' )" for col in columns])}].Name
-        """)
+        # We cannot rely on search to find directly columns as there might be Nulls. So intrating old fasion.
```
iakov-aws, line 607:
```suggestion
        # We cannot rely on search to find directly columns as there might be Nulls. So iterating old fashion.
```

cid/helpers/athena.py
```diff
@@ -595,7 +595,7 @@ def find_tables_with_columns_in_information_schema(self, columns):
         ''')
 
     def find_tables_with_columns(self, columns: list, database_name: str=None, catalog_name: str=None, max_items: int=10000):
-        """ This function searches a table with a given set of columns
+        """ This function return interator of tables with a given set of columns.
```
petrokashlikov, line 598: I think there is typo here and overall description kind of doesn't make sense.

iakov-aws: 1. function returns interator
2. iterrator that iterates tables 
3. but not all. only those that have a given set of columns 
 
 any suggestion how to phase it better?

iakov-aws: gen ai suggestion: 

Returns an iterator that yields only tables containing all specified columns.

iakov-aws:
```suggestion
        """ Returns an iterator that yields only tables containing all specified columns.
```

PR #1146, opened 2025-03-19
updating trends sql to use account_map

oladuni-aws:
*Issue #, if available:*

*Description of changes:*
replaced the reliance on aws_accounts and payer_account_name_map with account_map.
the goal should be just to update the script and not require extra steps to edit the quicksight dataset.

### DIFF
--- a/cid/builtin/core/data/queries/trends/monthly_bill_by_account.sql
+++ b/cid/builtin/core/data/queries/trends/monthly_bill_by_account.sql
@@ -19,15 +19,10 @@ WITH
    SELECT
      "account_name"
    , "account_id"
+   , "parent_account_name" "payer_account_name"
+   , "parent_account_id"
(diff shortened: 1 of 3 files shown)


cid/builtin/core/data/queries/trends/monthly_bill_by_account.sql
```diff
@@ -19,15 +19,10 @@ WITH
    SELECT
      "account_name"
    , "account_id"
+   , "parent_account_name" "payer_account_name"
+   , "parent_account_id"
```
iakov-aws, line 23: account_map does not have to contain parent_account_id or parent_account_name

iakov-aws: and this query will break in this case

PR #1194, opened 2025-04-25
Fix Glue database for CUR2

patrickmagyarskaylink:
*Issue #, if available:*
/

*Description of changes:*

When using CUR2, two databases are created in Glue. One from the https://aws-managed-cost-intelligence-dashboards.s3.amazonaws.com/cfn/data-exports-aggregation.yaml
Template: the cid_data_export database:
```
CIDDatabase:
Type: AWS::Glue::Database
Condition: DeployAnyTable
Properties:
DatabaseInput:
Name: !Join [ '_', !Split [ '-', !Sub '${ResourcePrefix}_data_export' ] ] # replace '-' to '_'
CatalogId: !Sub "${AWS::AccountId}"
```

with the associated Glue crawlers.
And once again, from the cid-cfn.yml template (https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/blob/main/cfn-templates/cid-cfn.yml#L1026), the "cid_cur" database but without the associated crawler: https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/blob/main/cfn-templates/cid-cfn.yml#L1031, because the condition "NeedCURTable" is not true, as it depends on the condition "NeedLegacyCUR," and this is also not true only for CUR2.
In Quicksight, the datasets are then created based on the incorrectly created database "cid_cur," which contains no data, and the dashboard therefore does not display anything.

The pull request changes the condition "NeedDatabase" which means that the database "cid_data_export" is used consistently.

### DIFF
--- a/cfn-templates/cid-cfn.yml
+++ b/cfn-templates/cid-cfn.yml
@@ -263,7 +263,7 @@ Parameters:
     Type: String
     Description: 'Currency Symbol. Please align with Currency of your CUR. Only supported for China region.'
     Default: "USD"
-    AllowedValues: 
+    AllowedValues:
       # Only currency symbols supported by QS
       - "USD" # US Dollar         $
       - "GBP" # British Pound     £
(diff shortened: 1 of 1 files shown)


iakov-aws:
Yes 2 databases are intentional design. 

data exports is just one of many sources CID supports. Even more will come. 
also different dashboards can require combination of several data sources and joins in various ways o data from different databases. 

This makes us choose this design that data exports is delivered with one database, data-collection creates its own database, etc. 

 and then Cid stores its views in a dedicate databases as well. 
 
 please let me know if any issue for you with this design.

iakov-aws:
As discussed in the call this works without this pr. So closing. Super happy to discsuss if any other feedback or enhancements

PR #1196, opened 2025-05-03
Handle paginated results from athena queries

joyofhex:
*Issue #, if available:*

*Description of changes:*

When athena has a large result set we run into the page size of 1,000 items. This PR changes get_query_results to use a paginator and depaginate the result set to support those situations.


iakov-aws (approved):

iakov-aws:
great catch!

PR #1231, opened 2025-06-11
Add support for AWS Athena managed query results

NithinChandranR-AWS:
*Issue #:* *https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/issues/1230*

## Summary

This PR adds support for the new AWS Athena managed query results feature to the CID framework. AWS launched this feature in June 2025 - it automatically manages query result storage at no additional cost, eliminating the need for customers to provision and manage S3 buckets.

IMG_8898

### CloudFormation template
- Added `AthenaQueryResultsMode` parameter (defaults to `customer-managed` for backward compatibility)
- Updated workgroup configuration to conditionally use managed results
- Added proper KMS encryption support for managed mode
- Kept all existing S3 bucket logic for customer-managed mode

### Python code  
- Modified `_ensure_workgroup()` in `cid/helpers/athena.py` to handle both modes
- Added parameter detection to determine which mode to use
- Updated workgroup creation to work without OutputLocation when using managed mode

## Why this change?

The main benefits are cost savings and operational simplification:
- Eliminates S3 storage costs for query results (can save $0.69-$69/month depending on usage)
- No need to provision or manage S3 buckets for query results
- Results are automatically cleaned up after 24 hours
- Enhanced security with workgroup-scoped access instead of S3 bucket permissions

## Multi-cloud considerations

For customers using FOCUS dashboards with multi-cloud data, we've kept the default as `customer-managed` mode. The managed mode is completely optional - customers can choose based on their requirements. No existing deployments will be affected.

## Encryption support

Managed query results support both AWS owned keys and customer managed keys (CMK), same as the current customer-managed approach. The CloudFormation template properly configures the required service principal for CMK usage.

## Testing

- Tested both managed and customer-managed modes
- Verified parameter flow from CloudFormation to Lambda
- Confirmed workgroup creation works in both scenarios
- Need to do more testing with large datasets and QuickSight integration

## Backward compatibility

This is fully backward compatible. The default parameter value is `customer-managed`, so existing deployments continue working exactly as before. Customers can opt into managed mode by changing the parameter.

## Files changed

- `cfn-templates/cid-cfn.yml` - Added parameter and conditional workgroup configuration
- `cid/helpers/athena.py` - Updated workgroup creation logic

### DIFF
--- a/ATHENA_MANAGED_QUERY_RESULTS_IMPLEMENTATION.md
+++ b/ATHENA_MANAGED_QUERY_RESULTS_IMPLEMENTATION.md
@@ -1,0 +1,154 @@
+# AWS Athena Managed Query Results Implementation
+
+## Overview
+Successfully implemented support for AWS Athena managed query results in the Cloud Intelligence Dashboards (CUDOS) framework. This feature was launched by AWS recently and eliminates the need for customer-managed S3 buckets for storing Athena query results.
+
+## Changes Made
+
+### 1. CloudFormation Template Changes (`cfn-templates/cid-cfn.yml`)
+
+#### New Conditions Added:
+- `UseAthenaManaged`: Determines if managed query results should be used
+- Updated `NeedAthenaQueryResultsBucket`: Only creates S3 bucket for customer-managed mode
+
+#### Parameter Updates:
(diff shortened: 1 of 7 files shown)


iakov-aws:
i will close this PR untill QS supports this feature

PR #1276, opened 2025-07-28
Change the terraform configurations to support as a module

meraj-kashi:
# Refactor Terraform Configuration for Module Reusability

This PR refactors the Terraform configuration to improve its usability as a module within other projects.

Currently, the Terraform module includes a hardcoded backend configuration and AWS provider definitions, which makes it difficult to integrate the module into other Terraform-based infrastructures.

The final goal is to consume the module in the below way:

```
module "cloud-intelligence-dashboard" {
  source = "github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework//terraform/cicd-deployment?ref=<release-tag>"

}
```

## Challenges Addressed

1. The Terraform **backend configuration** should not be included in the module itself. It must be defined by the consumer project to maintain flexibility.
2. An **IAM role for the destination account** was hardcoded, preventing deployments where the destination account is the primary executor (i.e., when not using cross-account roles). This setup assumes a cross-account deployment model from the payer (master) account only.

## Summary of Changes

1. **Removed** the backend configuration from the module.
2. **Removed** AWS provider definitions from `provider.tf`.
3. **Added** `configuration_aliases` to the `terraform` block to support both payer and destination account deployments.
4. **Updated** the `README.md` with new instructions for using the module, including deployment details and the rationale for these changes.
5. **Modified** `deploy.sh` to generate a `versions.tf` file containing the AWS provider configuration.  
   Previously, the script generated a `_overrides.tf` file, which conflicted with the current provider setup.


iakov-aws:
@vtaws 
please can you check that?

abportela:
+1 just to increase the pressure here :)

This is blocking our adoption of CID.

abportela:
@vtaws any updates about this?

sophie-warner-annalise-ai:
Would it be possible for this to be merged in the meantime and then overwritten when the AWS provided solution is ready?

I want to add a new integration to my CUR which is currently on legacy. I don't want to write the queries for the integration with legacy and I'm also not interested in the CLI tool or CF stacks.

Might just use the branch version of this: https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-data-collection/pull/293/files
Unsure about the relationship between the two repos.

PR #1323, opened 2025-09-11
Media Services Insights Hub v1.0.0 release

esc1144:
*Issue #, if available:*

*Description of changes:*
First production version of the MSIH dashboard


dashboards/media-services-insights/msih.yaml
```diff
@@ -0,0 +1,816 @@
...
+           "elemental_service"
+         , array_agg("pattern_exp" ORDER BY priority ASC) "patterns"
+         , array_agg("replace_exp" ORDER BY priority ASC) "reps"
+         FROM
+           "emc_usage_translation_rules"
+         GROUP BY 1
+      )
+      , "cur_pre_processing" AS (
+         SELECT
+           *
```
yprikhodko, line 723: If CUR will get additional columns such as SCAD or others this Athena query may fail with error the view is stale. Any reasons why not to list columns explicitly there from which used in the resulting select?

PR #1398, opened 2025-12-30
Feature/comprehensive tags column support

brennopassanha:
# Add comprehensive Tags column support for CUR 2.0

## Summary
This PR implements support for the new comprehensive `tags` column introduced in AWS Cost and Usage Report (CUR) 2.0, specifically focusing on Account Tags functionality. The implementation enables users to access Account Tags in CID.

### Account Tags Support from Comprehensive Tags Column
- **New Account Tags functionality**: Extracts account-level tags from the comprehensive `tags` column in CUR2
- **Filtered extraction**: Uses `accountTag/` prefix filtering to isolate account tags from the comprehensive column
- **Prefixed naming**: Account tags are prefixed with `account_tag_` to distinguish them from resource tags
- **Backward compatibility**: Maintains existing support for `resource_tags` and `cost_category` columns

### Enhanced Tag Discovery Display
- **Top 20 display**: Shows the 20 most useful tags by cardinality, removing the previous >10 filter limitation
- **Better formatting**: 
  - Clear section headers showing "(Top 20)" 
  - "Account Tags" label instead of generic "tags"
  - Empty lines between sections for improved readability
- **Inclusive approach**: Displays tags regardless of cardinality count, better supporting smaller accounts and test environments
- **Consistent sorting**: All tag types sorted by distinct values count (highest first)

### Improved Error Handling
- **Empty result safety**: Prevents crashes when tag discovery returns no results
- **Graceful degradation**: Continues processing even when specific tag columns are missing
- **Informative messages**: Clear error messages explaining missing column scenarios

## Technical Implementation
- Maintains separate scanning of `resource_tags`, `cost_category`, and `tags` columns
- Updates column type detection to handle the comprehensive `tags` column as MAP type
- Preserves all existing functionality while adding new Account Tags capabilities
- No breaking changes to existing CUR table support

This enables Account Tags functionality in QuickSight datasets and dashboards for users with CUR2 tables that include the comprehensive tags column.

### DIFF
--- a/cid/helpers/cur.py
+++ b/cid/helpers/cur.py
@@ -190,7 +190,7 @@ class AbstractCUR(CidBase):
                 if tag_type not in self.fields:
                     logging.debug(f'skipping {tag_type} scan')
 
-                display_name = "Account Tags" if tag_type == "tags" else tag_type
+                display_name = "Account Tags (tags column)" if tag_type == "tags" else tag_type
                 cid_print(f'Scanning {display_name} in {self.table_name}.')
                 try:
                     if tag_type == 'tags':
--- a/cid/helpers/quicksight/dashboard_patching.py
+++ b/cid/helpers/quicksight/dashboard_patching.py
@@ -57,10 +57,13 @@ def format_field_name(field_name: str, ignore_prefix: bool=False) -> str:
             i += 1
     title = ' '.join(result)
     title = title.replace('Aws ', 'AWS ')
+    # replace prefixes to make filter titles shorter on the dashboards. We don't do the same replacement for group by patching to avoid name overlap there 
     if ignore_prefix:
         new_title = title
         if title.startswith('Cost Category '):
             new_title = title.replace('Cost Category ', '')
+        elif title.startswith('Account Tag '):
+            new_title = title.replace('Account Tag ', '')
         elif title.startswith('Tag '):
             new_title = title.replace('Tag ', '')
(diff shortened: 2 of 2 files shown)


cid/common.py
```diff
@@ -1931,15 +1910,18 @@ def _tag_to_name(tag):
...
             )
             if not tag_name.startswith('tag_'):
                 if tag.startswith('cost_category'):
                     tag_name = 'cost_category_' + tag_name
+                elif tag.startswith('tags'):
+                    tag_name = 'account_tag_' + tag_name
                 else:
                     tag_name = 'tag_' + tag_name
-            return tag_name.replace(':', '_')
+            return re.sub(r'\W', '_', tag_name)
```
yprikhodko: @brennopassanha why is this change needed?

brennopassanha: This is replacing all special characters from all three columns (tags, resource_tags and cost_category) with underscore.
Before, we were replacing only colons, because resource_tags and cost_category values are normalized by default, for example:
A resource tag `Team-Brennop` is pushed to resource_tags column as `user_team_brennop`; But in the tags column, it appears as `user:Team-Brennop`

Quick Suite does not support special characters on calculated fields (where these values end up at), that's the reason colons were replaced before. But now with the addition of the non-normalized tags column, we have to replace other special characters - and I thought an easy win here was to replace any special character it may exist with underscore.

PR #1443, opened 2026-03-23
Fix: Add tags_json to msih_view and make tag columns null-safe

ihartzaws:
The Media Services Insights dashboard deployment fails with CONTEXTUAL_UNKNOWN_SYMBOL error on tag_cost_center because the Athena view doesn't include tags_json, which the QuickSight dataset definition depends on.

Changes:

1. msih_view SQL — Added tags_json column derived from CUR2 resource_tags MAP, using json_format for Athena VARCHAR compatibility: sql IF(cardinality(MAP_FILTER(resource_tags, (k, v) -> v IS NOT NULL AND v <> '')) > 0, json_format(CAST(MAP_FILTER(resource_tags, (k, v) -> v IS NOT NULL AND v <> '') AS JSON)), NULL) "tags_json"
   

2. QuickSight dataset definition — Made tag_cost_center and tag_business_unit calculated columns null-safe to handle customers with no tags:
  
   ifelse(isNull({tags_json}) OR {tags_json} = '{}', NULL, parseJson({tags_json}, '$.tag_cost_center'))
   

Root cause:

The Athena view (msih_view) did not include tags_json in its SELECT, but the QuickSight dataset unconditionally referenced it via parseJson, causing CreateDataSet to fail for any customer where the view lacked tag data.

Notes:
- Empty tag maps return NULL instead of '{}' to avoid Athena cast errors on empty JSON objects
- Athena views don't support JSON column types, so json_format() is used to output VARCHAR
- The null-safe ifelse in QuickSight handles both NULL and empty '{}' cases

Testing:

Verified against a customer deployment where the original error occurred. Dashboard deploys and SPICE syncs successfully after the fix.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/dashboards/media-services-insights/msih.yaml
+++ b/dashboards/media-services-insights/msih.yaml
@@ -593,12 +593,12 @@ datasets:
               Columns:
               - ColumnName: tag_cost_center
                 ColumnId: 1971599a-c062-5962-8087-aa1e44db949c
-                Expression: parseJson(tags_json, '$.tag_cost_center')
+                Expression: ifelse(isNull({tags_json}) OR {tags_json} = '{}', NULL, parseJson({tags_json}, '$.tag_cost_center'))
           - CreateColumnsOperation:
               Columns:
               - ColumnName: tag_business_unit
                 ColumnId: c9a78250-5c66-57f0-ac6f-5c9ac7344b8d
-                Expression: parseJson(tags_json, '$.tag_business_unit')
+                Expression: ifelse(isNull({tags_json}) OR {tags_json} = '{}', NULL, parseJson({tags_json}, '$.tag_business_unit'))
           - CreateColumnsOperation:
               Columns:
               - ColumnName: linked_account_formatted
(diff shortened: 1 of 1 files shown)


ihartzaws:
Update:

The Media Services Insights dashboard deployment fails with CONTEXTUAL_UNKNOWN_SYMBOL error on tag_cost_center because the Athena view doesn't include tags_json, which the QuickSight dataset definition depends on. Additionally, the 
CreateColumnsOperation expressions for tag columns don't handle null values, causing failures for customers without tags.

Changes:

1. msih_view SQL — Added tags_json column derived from CUR2 resource_tags MAP, using json_format for Athena VARCHAR compatibility:
  sql
   IF(cardinality(MAP_FILTER(resource_tags, (k, v) -> v IS NOT NULL AND v <> '')) > 0,
     json_format(CAST(MAP_FILTER(resource_tags, (k, v) -> v IS NOT NULL AND v <> '') AS JSON)),
     NULL) "tags_json"
   

2. QuickSight CreateColumnsOperation for tag_cost_center and tag_business_unit — The original Expression in both CreateColumnsOperation blocks did not handle null tags_json values, causing dataset creation to fail for customers without tags. 
Changed from:
  
   Expression: parseJson(tags_json, '$.tag_cost_center')
   
  To:
  
   Expression: ifelse(isNull({tags_json}) OR {tags_json} = '{}', NULL, parseJson({tags_json}, '$.tag_cost_center'))
   

Notes:
- Empty tag maps return NULL to avoid Athena cast errors on empty JSON objects
- Athena views don't support JSON column types, so json_format() is used to output VARCHAR
- The null-safe ifelse in the CreateColumnsOperation expressions handles both NULL and empty '{}' cases

Testing:

Verified against a customer deployment where the original errors occurred. Dashboard deploys and SPICE syncs successfully after the fix.

PR #1457, opened 2026-04-08
Feature/account mapping

martrost:
# cid-cmd map — Account Mapping

The `map` command creates an `account_map` Athena view that enriches your AWS account data with custom taxonomy dimensions (business unit, environment, cost center, etc.). These dimensions can then be used in Cloud Intelligence Dashboards for grouping, filtering, and reporting.

## How It Works

The command follows a six-phase workflow:

1. **Discovery** — auto-detects the `organization_data` table (from AWS Organizations data collection) and the target Athena database
2. **Configuration** — prompts you to select data sources and define taxonomy dimensions (or reuses a saved configuration)
3. **Data Loading** — reads organization data from Athena and optionally loads an external file
4. **Transformation** — applies taxonomy rules to produce the enriched account map
5. **Preview** — shows a sample of the output and the generated SQL for confirmation
6. **Write** — creates the Athena views (`account_map`, `account_map_config`, and optionally `account_map_file_source`)

## Usage

```bash
# Interactive mode (default) — walks you through configuration
cid-cmd map

# Provide a file with additional taxonomy data
cid-cmd map --file accounts.csv

# Legacy mode — simple account_id/account_name mapping from organization_data
cid-cmd map --simple

# Custom output view name
cid-cmd map --view-name my_account_map
```

## Options

| Option | Description |
|---|---|
| `--file PATH` | Path to a CSV, Excel, or JSON file containing additional taxonomy columns to join by account ID |
| `--simple` | Use legacy mode — creates a basic `account_map` view with just `account_id` and `account_name` from the organization_data table |
| `--view-name TEXT` | Name of the output Athena view (default: `account_map`) |
| `-v` / `--verbose` | Increase log verbosity (can be repeated: `-vv`) |
| `-y` / `--yes` | Auto-confirm all prompts |

## Taxonomy Dimension Sources

During interactive configuration you choose one or more data sources for your taxonomy dimensions:

### Tags from source table
Extracts values from the `hierarchytags` column in `organization_data`. Each selected tag key becomes a column in the output view.

Example: if your accounts are tagged with `Environment=Production` and `CostCenter=Engineering`, selecting those tag keys produces `environment` and `cost_center` columns.

### Additional file (`--file`)
Joins columns from a CSV/Excel/JSON file by account ID. The file must contain an account ID column; all other selected columns become taxonomy dimensions.

Example file (`accounts.csv`):
```csv
account_id,business_unit,team
123456789012,Retail,Frontend
234567890123,Platform,Data
```

### Split account name
Extracts dimensions by splitting the `account_name` string on a separator character. You specify the separator and the positional index to extract.

Example: for account name `aws-retail-prod`, splitting by `-` at index 1 yields `retail`, and at index 2 yields `prod`.

## Configuration Persistence

The command saves its configuration as an Athena view (`<view_name>_config`). On subsequent runs it detects the existing config and offers to reuse it, so you don't have to reconfigure every time.

## Views Created

| View | Purpose |
|---|---|
| `account_map` | The main enriched account mapping view used by dashboards |
| `account_map_config` | Stores the mapping configuration for reuse |
| `account_map_file_source` | (Only when `--file` is used) Stores the file data as an Athena view for joins |

## Prerequisites

- An `organization_data` table in Athena (typically created by the CID data collection CFN stack)
- Athena workgroup with a configured query result location
- Appropriate IAM permissions for Athena and Glue operations

## Examples

Create an account map using AWS Organization tags:
```bash
cid-cmd map
# → Select "Tags from source table"
# → Pick tag keys like Environment, CostCenter, Team
# → Preview and confirm
```

Create an account map enriched with data from a spreadsheet:
```bash
cid-cmd map --file ~/Downloads/account_taxonomy.xlsx
# → Select "Additional file" and/or "Tags from source table"
# → Pick columns from the file to use as dimensions
# → Preview and confirm
```

Re-run with saved configuration (no prompts):
```bash
cid-cmd map -y
```


martrost:
Added the docs/cid-cmd update so the documentation is available as well

requirements.txt
```diff
@@ -6,4 +6,4 @@ six>=1.15
 tqdm
 tzlocal>=4.0
 InquirerPy
-filelock>=3.20.3
+pandas>=2.3.5
```
yprikhodko, line 9: why do we need this?

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+            elif source_type == 'file':
+                if file_source_view:
+                    dimension_parts.append((output_name, f"file.{source_value} AS {output_name}"))
+                else:
+                    logger.warning(f"File source dimension {dim_name} specified but no file source view")
+                    dimension_parts.append((output_name, f"NULL AS {output_name}"))
+                    
+            elif source_type == 'name_split':
+                if isinstance(source_value, dict):
+                    separator = source_value.get('separator', '-')
```
yprikhodko, line 3036: Validate/sanitize the separator field (e.g., restrict to 1-2 characters, alphanumeric or common delimiters only) or escape it properly.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+    
+    Attributes:
+        config (dict): Application configuration
+        athena_helper (Athena): CID Athena helper instance
+        MAX_SQL_SIZE (int): Maximum SQL statement size in bytes (262144)
+    """
+    
+    MAX_SQL_SIZE = 262144  # Athena's maximum SQL statement size in bytes
+    
+    def __init__(self, config: dict, athena_helper: Athena = None):
```
yprikhodko: why not to make athena_helper required and eliminate need for the checks below?

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+        _payer_account_id (str): Payer account ID (12 digits)
+        _payer_account_name (str): Payer account name
+        _account_tags (dict): Account tags
+        _business_unit (dict): Business unit taxonomy information
+        
+    Class Attributes:
+        _all_accounts (dict): Class-level storage of all Account instances
+    """
+    
+    _all_accounts: Dict[str, Dict] = {}
```
yprikhodko: Class-level mutable state (_all_accounts) is a shared-global anti-pattern — it accumulates data across all instances and persists between calls within the same process. If reset_all_accounts() isn't called (e.g., an error occurs before TransformEngine.transform() reaches it), stale data leaks into subsequent runs.

Consider replacing this with an instance-scoped AccountRegistry owned by TransformEngine, so each transformation run gets its own isolated storage with no risk of cross-run contamination.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+        """
+        Get all Account instances and their metadata.
+        
+        Returns:
+            Dictionary with account IDs as keys and metadata dictionaries as values
+        """
+        return cls._all_accounts
+    
+    @classmethod
+    def reset_all_accounts(cls) -> None:
```
yprikhodko: The need for this method is a code smell — it exists solely to work around the class-level mutable state. If the registry were instance-scoped (owned by TransformEngine), cleanup would happen automatically via garbage collection when the engine goes out of scope.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+        Returns:
+            DataFrame containing account map with all taxonomy dimensions
+            
+        Raises:
+            ValueError: If configuration is invalid or required data is missing
+        """
+        logger.info("Starting transformation process")
+        
+        # Reset Account class storage
+        Account.reset_all_accounts()
```
yprikhodko: This reset is the only thing preventing data leakage between runs, but it's fragile — if an exception occurs after accounts are created but before the next transform() call resets them, the stale data persists. Moving to an instance-scoped registry eliminates this concern entirely.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
+import json
```
yprikhodko, line 1: json is imported here at module level, but it's also imported again locally inside parse_config_rows() and generate_config_rows(). Remove the local import json statements inside those methods — the top-level import is sufficient and the redundant imports add confusion about whether there's a scoping reason for them.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+from InquirerPy.separator import Separator
+
+from cid.helpers import Athena
+from cid.exceptions import CidCritical
+
+logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def spinner(message: str = "Processing..."):
```
yprikhodko, line 35: This spinner writes to sys.stdout from a background thread while the main thread may also be writing (e.g., InquirerPy prompts, print() calls). This can produce garbled terminal output.
Consider clearing the spinner line (\r + spaces) before yielding back to the caller or at minimum, add a sys.stdout.write("\r" + " " * (len(message) + 4) + "\r") before the yield to ensure the line is clean when the caller starts writing

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+
+        results = {
+            'file_source_view': None,
+            'config_view': None,
+            'account_map_view': None,
+            'deleted_views': []
+        }
+
+        # Get view name from config or use default
+        view_name = getattr(self, 'view_name', 'account_map')
```
yprikhodko: self.view_name is never set in AthenaWriter.__init__, so this always falls through to the default 'account_map'. The subsequent line tries self.config.get('general', {}).get('view_name', ...) which also won't match the actual config structure used in this PR.

Fix: either accept view_name as an __init__ parameter and store it as self.view_name, or just use the view_name argument that's already passed into write_complete_mapping directly (which is what happens on the next line via the method parameter).

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+        for db in databases:
+            try:
+                metadata = self.athena.get_table_metadata('organization_data', db)
+                if metadata:
+                    matches.append(db)
+            except Exception:
+                continue
+
+        if len(matches) == 1:
+            logger.info(f"Found 'organization_data' in database '{matches[0]}' — auto-selected")
```
yprikhodko: The existing codebase uses %s-style logging throughout (e.g., logger.info("Loaded %d accounts from Athena", len(df))). f-string logging eagerly evaluates the string even when the log level is disabled, which is a minor performance hit but more importantly breaks consistency.

This pattern appears in many places in this file . Please use logger.info("Found 'organization_data' in database '%s' — auto-selected", matches[0]) style consistently.

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+
+        Args:
+            config: Existing configuration to display
+
+        Returns:
+            bool: True if user wants to reuse config, False otherwise
+        """
+        from InquirerPy import inquirer
+
+        print("\n" + "="*60)
```
yprikhodko: The existing codebase uses cid_print from cid.utils for user-facing output (it handles encoding, piping, and non-TTY environments gracefully). Raw print() can fail in environments where stdout encoding doesn't support the emoji characters used here (✅, ❌, ⚠️, 📋, etc.).

Consider using cid_print for all user-facing output, or at minimum wrap the emoji-heavy output in a try/except for UnicodeEncodeError

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+
+        Args:
+            config: Configuration dictionary
+            df: Transformed account map DataFrame
+            database: Target database name
+
+        Returns:
+            dict: Results containing created view names and status
+        """
+        from InquirerPy import inquirer
```
yprikhodko: InquirerPy is imported with different aliases across this file:

from InquirerPy import inquirer (top-level, line 12)
from InquirerPy import inquirer as inq

Pick one and use it consistently. Since the top-level import already uses inquirer, just use that everywhere and remove the local re-imports with the inq alias

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+        # Validate preferred database if provided
+        if preferred:
+            if self.athena.get_database(preferred):
+                logger.info(f"Using preferred database: {preferred}")
+                return preferred
+            else:
+                logger.warning(f"Preferred database '{preferred}' not found, discovering alternatives...")
+        
+        # Get list of available databases
+        databases = self.athena.list_databases()
```
yprikhodko, line 1376: This method iterates through every database in the catalog and calls get_table_metadata('organization_data', db) for each one. In accounts with many Glue databases (50+), this generates N API calls sequentially and can be slow.

Consider:

Adding a short-circuit: check common database names first (optimization_data, cur, the default CID database) before falling back to a full scan
Adding a --database CLI option to skip discovery entirely
Caching the result so subsequent calls in the same session don't re-scan

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+            choices: List of choices
+            **kwargs: Additional kwargs passed to inquirer.checkbox
+
+        Returns:
+            list: Selected items (may be empty if user confirms)
+        """
+        from InquirerPy import inquirer as inq
+
+        while True:
+            result = inq.checkbox(
```
yprikhodko: f this code runs in a non-interactive environment, InquirerPy prompts will raise an exception or hang
Consider checking cid.utils.isatty() at the start of the interactive workflow and raising a clear error like:

if not isatty():
    raise CidCritical(
        "Interactive mode requires a TTY. Use 'cid-cmd map --simple' for non-interactive environments."
    )

cid/helpers/account_mapper_helpers.py
```diff
@@ -0,0 +1,3329 @@
...
+            
+            success_count = 0
+            for account_id in all_accounts.keys():
+                try:
+                    # Apply the rule for this dimension
+                    value = self.apply_single_rule(dimension_config, account_id)
+                    
+                    if value is not None:
+                        # Create Account instance to set the business unit
+                        account = Account(account_id)
```
yprikhodko: This creates a new Account instance for an account that already exists in _all_accounts (it was created during the earlier loop in transform()). The constructor silently reuses the existing entry in the class-level dict, but reading this code it looks like a new account is being created for each dimension rule application.

This is confusing and fragile — it only works because of the class-level state side effect. If you keep the current architecture, at minimum rename this to something like Account.get_or_create(account_id) to make the intent clear. Better yet, this is another reason to move to the instance-scoped registry pattern where you'd write:



self.registry.set_field(account_id, dimension_name, value)

yprikhodko (changes requested):

requirements.txt
martrost: Pandas was used to facilitate the construction of output table and handle input files with csv, xlsx & json format. Removed Pandas, dropping xlsx, xls and json in the process (can be added back if customers need it down the line)

cid/helpers/account_mapper_helpers.py
martrost: Replaced input with defined list to select from (1st layer) and sanitation when processing (2nd layer)

cid/helpers/account_mapper_helpers.py
martrost: Done

cid/helpers/account_mapper_helpers.py
martrost: Adressed by replacing it with an instance-scoped registry

cid/helpers/account_mapper_helpers.py
martrost: Adressed by replacing it with an instance-scoped registry

cid/helpers/account_mapper_helpers.py
martrost: Missed that one, now using cid_print

cid/helpers/account_mapper_helpers.py
martrost: Aligned

cid/helpers/account_mapper_helpers.py
martrost: Good idea, implemented accordingly

cid/helpers/account_mapper_helpers.py
martrost: Addressed as well

cid/helpers/account_mapper_helpers.py
martrost: Adressed by replacing it with an instance-scoped registry

cid/helpers/account_mapper_helpers.py
martrost: Adressed by replacing it with an instance-scoped registry

cid/helpers/account_mapper_helpers.py
martrost: Done

cid/helpers/account_mapper_helpers.py
martrost: Fixed. The spinner thread now only writes animation frames. The completion message is written on the main thread in the finally block after the spinner thread is joined. The line is cleared with \r + spaces + \r before writing the completion message, ensuring a clean terminal state for whatever comes next.

cid/helpers/account_mapper_helpers.py
martrost: Removed the dead getattr/config.get('general') logic entirely. view_name is now an explicit parameter on write_complete_mapping(), and the caller (UnifiedWorkflow.execute) passes self.view_name directly. No more guessing from config or falling through to defaults.

cid/helpers/account_mapper_helpers.py
martrost: Adjusted to use the same logging style

yprikhodko (approved):

PR #1463, opened 2026-04-17
terraform support for remote data exports

yprikhodko:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/terraform/cicd-deployment/locals.tf
+++ b/terraform/cicd-deployment/locals.tf
@@ -38,7 +38,7 @@ locals {
         RolePath             = var.cid_dataexports_destination.role_path
         CUR2TimeGranularity  = var.cid_dataexports_destination.cur2_time_granularity
         FOCUSTimeGranularity = var.cid_dataexports_destination.focus_time_granularity
-        LegacyLocalBucket    = "no"
+        LegacyLocalBucket    = var.legacy_local_bucket
       }
       timeouts = local.default_timeouts
       tags     = local.common_tags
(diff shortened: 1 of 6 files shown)


terraform/cicd-deployment/locals.tf
```diff
@@ -38,6 +38,7 @@ locals {
         RolePath             = var.cid_dataexports_destination.role_path
         CUR2TimeGranularity  = var.cid_dataexports_destination.cur2_time_granularity
         FOCUSTimeGranularity = var.cid_dataexports_destination.focus_time_granularity
+        LegacyLocalBucket    = "no"
```
vtaws: Where did you add this parameter ? data-exports-aggregation.yaml ?

terraform/cicd-deployment/locals.tf
```diff
@@ -58,6 +59,7 @@ locals {
         RolePath             = var.cid_dataexports_source.source_role_path
         CUR2TimeGranularity  = var.cid_dataexports_source.source_cur2_time_granularity
         FOCUSTimeGranularity = var.cid_dataexports_source.source_focus_time_granularity
+        LegacyLocalBucket    = "no"
```
vtaws: Where did you add this parameter ? data-exports-aggregation.yaml ?

samarza: Also worth noting, for existing Terraform deployments this will trigger a CFN stack update and change LegacyLocalBucket from the template default "yes" to "no". Is that intentional, or should this be exposed as a variable defaulting to "yes" for back-compatibility?

terraform/cicd-deployment/locals.tf
samarza: Also worth noting, for existing Terraform deployments this will trigger a CFN stack update and change LegacyLocalBucket from the template default "yes" to "no". Is that intentional, or should this be exposed as a variable defaulting to "yes" for back-compatibility?

yprikhodko: yes in data-exports-aggregation.yaml

yprikhodko: It's exposed as parameter defaulting to yes for back compatibility. So expected behaviour:
1. Update -> it should use default value of yes and keep local S3 bucket as there is some data and it can't be deleted
2. New deployments -> it should be set to no and don't deploy local s3 bucket in source account

vtaws (approved):

PR #1468, opened 2026-04-29
Fix/ext supp taxonomy fields

juchavw:
*Issue #, if available:* https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/issues/1461

*Description of changes:*
fix: surface taxonomy fields on extended-support tagging view

The extended_support_tagging_view dataset wasn't joined to account_map, so its OutputColumns lacked the OU columns the other four datasets carried. The taxonomy picker intersects column names across datasets, so the missing columns collapsed the candidate list and the prompt was silently skipped.

- Project line_item_usage_account_id in the view SQL (and GROUP BY).
- Add the account_map LEFT join to the tagging view dataset, matching the pattern used by the other four datasets.
- Add account_map to dependsOn.views for the extended-support-tagging-view dataset definition.
- Bumped ext. supp. cost proj. dashboard version to 5.2.1 

### DIFF
--- a/dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
+++ b/dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
@@ -17,7 +17,16 @@ dashboards:
     - opensearch_extended_support_view
     - elasticache_extended_support_view
     nonTaxonomyColumns:
+    - account_email_id
+    - account_id
+    - account_name
+    - account_prefix
+    - account_suffix
+    - line_item_usage_account_id
     - line_item_product_code
+    - o_u
+    - parent_account_id
+    - parent_account_name
     - tags_json
     file: ./extended-support-cost-projection-definition.yaml
     version: v5.2.0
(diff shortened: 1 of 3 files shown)


dashboards/extended-support-cost-projection/extended-support-cost-projection.yaml
```diff
@@ -17,10 +17,19 @@ dashboards:
     - opensearch_extended_support_view
     - elasticache_extended_support_view
     nonTaxonomyColumns:
```
yprikhodko, line 19: Why exclude these?
account_email_id
account_prefix
account_suffix
o_u
parent_account_id
parent_account_name

They could be valid columns to add to the dashboard

juchavw: I've removed them from the exclusion list. I was thinking on not having columns that are already in the dashboard but they can be controlled with user selection.

yprikhodko (approved):

PR #1470, opened 2026-05-03
feat(cfn): Add PermissionsBoundary to all IAM roles

### LINKED ISSUE #1205: Enhancement request - Implement permissions boundaries on the CloudFormation template
We would like guidance on how to implement permissions boundaries on the CloudFormation template.

We believe that adding a permissions boundary outside of the CloudFormation template could cause issues when team needs to update via the template.

NithinChandranR-AWS:
Fixes #1205

## Summary
Add the missing PermissionsBoundary property to InitLambdaExecutionRole and PullAssetsLambdaRole. Also add the missing Path property to PullAssetsLambdaRole for consistency.

## Changes
The PermissionsBoundary parameter and NeedPermissionsBoundary condition already existed in the template and were applied to 5 of 7 IAM roles. This ensures all roles respect the boundary when specified, enabling deployment in environments with SCP-enforced boundary requirements.

## Testing
- Verified all 7 IAM roles now have PermissionsBoundary
- Default behavior unchanged (empty string = no boundary applied)

### DIFF
--- a/cfn-templates/cid-cfn.yml
+++ b/cfn-templates/cid-cfn.yml
@@ -1975,6 +1975,8 @@ Resources:
             Principal:
               Service: lambda.amazonaws.com
             Action: sts:AssumeRole
+      Path: !Ref RolePath
+      PermissionsBoundary: !If [NeedPermissionsBoundary, !Ref PermissionsBoundary, !Ref 'AWS::NoValue']
       ManagedPolicyArns:
         - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole'
       Policies:


yprikhodko (commented):
Thanks for the fix! The PullAssetsLambdaRole change (adding Path + PermissionsBoundary) looks correct and is needed.

However, it looks like this PR is based on an older version of main. InitLambdaExecutionRole already has PermissionsBoundary applied (added in a prior release). Could you rebase against current main and drop that hunk? Otherwise this will either conflict or produce a duplicate property.

NithinChandranR-AWS:
Good catch, thanks.  Rebased on current main and dropped the InitLambdaExecutionRole hunk — only PullAssetsLambdaRole remains now (Path + PermissionsBoundary).

NithinChandranR-AWS:
Thanks for the quick review, Yuriy! Rebased on current main and dropped the InitLambdaExecutionRole hunk -- only PullAssetsLambdaRole remains now. Force-pushed.

NithinChandranR-AWS:
removed Hunk 1 (the InitLambdaExecutionRole change) because it already exists upstream — kept only Hunk 2 (PullAssetsLambdaRole) . Resubmitted for review.

yprikhodko (approved):

PR #1491, opened 2026-06-01
added support of line_item_iam_principal column as taxonomy

yprikhodko:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -2021,6 +2021,8 @@ class Cid():
         ''' returns an sql for json tag
         '''
         def _tag_to_name(tag):
+            if tag == 'line_item_iam_principal':
+                return 'line_item_iam_principal'
             tag_name = (tag
                 .replace('resource_tags_', '')
                 .replace('cost_category_', '')
--- a/cid/common.py
+++ b/cid/common.py
@@ -2022,7 +2022,7 @@ class Cid():
         '''
         def _tag_to_name(tag):
             if tag == 'line_item_iam_principal':
-                return 'line_item_iam_principal'
+                return 'iam_principal'
             tag_name = (tag
                 .replace('resource_tags_', '')
                 .replace('cost_category_', '')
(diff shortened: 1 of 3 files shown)


esc1144 (approved):

cid/common.py
```diff
@@ -2037,7 +2039,7 @@ def _tag_to_name(tag):
                 elif "userAttribute/" in tag:
                     tag_name = 'user_attribute_' + tag_name
```
brennopassanha: Just like IAMPrincipal **Tags**, I would expect UserAttributes **Tags** to follow the same naming convention:

```
tag_name = 'user_attribute_tag_' + tag_name
```

cid/helpers/cur_proxy.py
```diff
@@ -51,6 +51,7 @@ def batched(iterable, n):
     'line_item_line_item_description': 'line_item_line_item_description',
     'line_item_tax_type': 'line_item_tax_type',
     'line_item_legal_entity': 'line_item_legal_entity',
+    'line_item_iam_principal': 'line_item_iam_principal',
```
brennopassanha: Why do we need to add line_item_iam_principal on anything related to cur proxy?
If an user is already on CUR2 and we are working on a Dashboard that requires CUR Legacy (not sure if this scenario exists), this column won't be used, because it doesn't exist on CUR Legacy anyway.

If a user is on CUR Legacy, we won't have this column to populate a proxy to CUR2.

It should be fine if this is only a placeholder, that will be casted as NULL anyway.

brennopassanha (approved):
It works well, approved.
Optionally, both comments I left can be considered for a change if author agrees with those.

cid/common.py
yprikhodko: UserAttributes are actually not a tags, but user attributes coming from SSO. That's why I decided not to add tag suffix there as customers actually don't tag SSO resources but just decide which attributes will appear in CUR

cid/helpers/cur_proxy.py
yprikhodko: Removed, you're right

PR #1500, opened 2026-06-10
fix(cur_proxy): Return SQL expression instead of None for CUR2→CUR1 proxy fields

### LINKED ISSUE #1498: PCA Dashboard incompatible with CUR 2.0 — cur_proxy.py returns Python None + missing columns in default Data Export
# GitHub Issue — CID PCA + CUR 2.0 Bug

**Paste at:** https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/issues/new

---

**Title:** PCA Dashboard incompatible with CUR 2.0 — cur_proxy.py returns Python None + missing columns in default Data Export

**Body:**

## Bug Description

Deploying the Pricing Change Analysis (PCA) dashboard using `cid-cmd v4.4.14` with a CUR 2.0-only account (no legacy CUR) fails with two cascading issues.

## Environment
- cid-cmd version: 4.4.14
- CUR type: CUR 2.0 Data Export (created via `data-exports-aggregation.yaml` CFN stack)
- No legacy CUR present
- Region: us-east-1
- Dashboards that DO work: CUDOS v5, Cost Intelligence, KPI (native CUR 2.0 support)

## Issue 1: `cur_proxy.py` returns Python `None` instead of SQL NULL

**File:** `cid/helpers/cur_proxy.py`, line ~536-543

When `current_cur_version='2'` and `target_cur_version='1'` (creating a CUR1 proxy from CUR2 data), `get_sql_expression()` has an incomplete code path:

```python
if self.current_cur_version.startswith('2') and self.target_cur_version.startswith('1'):
    for tag_type in 'resource_tags', 'cost_category':
        if field.startswith(tag_type + '_'):
            short_field_name = field[len(tag_type + '_'):]
            return f"{tag_type}['{short_field_name}']"
    # ← No return statement here — returns Python None
```

For fields like `month`, `year`, `product_cache_engine`, etc., the function returns `None` (implicit Python return). This gets interpolated directly into the SQL string as the literal word `None`, causing:

```
Athena query status failed: line 24:18: Column 'none' cannot be resolved
```

**Proposed fix:** Add a fallback return after the tag loop:
```python
    return empty.get(field_type.lower(), 'CAST(NULL AS varchar)')
```

## Issue 2: PCA requires columns not included in default CUR 2.0 Data Export

Even after manually fixing the None bug, PCA's `pricing_changes` view fails because it references columns that don't exist in the Glue table created by `data-exports-aggregation.yaml`:

- `product_sku`
- `product_product_name`
- `product_region`
- `pricing_public_on_demand_rate`
- `pricing_rate_code`
- `line_item_unblended_rate`
- `line_item_net_unblended_cost`

These columns exist in the CUR 2.0 schema but are not selected in the default Data Export configuration created by the CFN template.

Error:
```
Athena query status failed: line 10:12: Column 'product_sku' cannot be resolved
```

## Impact

**All users** who:
1. Use CUR 2.0 only (no legacy CUR)
2. Deploy CID via the standard CFN process
3. Attempt to add PCA dashboard

Will hit both issues. PCA is effectively broken for CUR 2.0-only customers.

## Steps to Reproduce

1. Deploy `data-exports-aggregation.yaml` (Step 1 from docs)
2. Deploy `cid-cfn.yml` (Step 3 — CUDOS/KPI/CI work fine)
3. Run `cid-cmd deploy --dashboard-id pricing-change-analysis`
4. Select `cid_data_export.cur2` as CUR table
5. Observe `None` error on cur1_proxy creation

## Suggested Fix

1. **cur_proxy.py**: Add explicit return for the CUR2→CUR1 path when field is not a tag/cost_category
2. **data-exports-aggregation.yaml**: Include all columns required by PCA in the CUR 2.0 export SQL, or document that PCA requires a custom export configuration
3. **Alternatively**: Update PCA dashboard to use CUR 2.0 natively (like CUDOS v5 does) instead of relying on cur1_proxy

## Related

- PR #1423 (may partially address this)

NithinChandranR-AWS:
## Problem

get_sql_expression() in cur_proxy.py has an incomplete code path when current_cur_version='2' and target_cur_version='1'. For non-tag fields (e.g. month, year, product_sku), the function falls through without a return, producing Python None which gets interpolated as literal 'None' in SQL.

PCA dashboard is broken for CUR 2.0-only customers.

## Fix

After the tag/cost_category loop, look up the field in cur1to2_mapping (CUR1→CUR2 SQL equivalents). If not found, return typed NULL.

Fixes #1498

### DIFF
--- a/cid/helpers/cur_proxy.py
+++ b/cid/helpers/cur_proxy.py
@@ -539,6 +539,11 @@ class ProxyView():
                 if field.startswith(tag_type + '_'):
                     short_field_name = field[len(tag_type + '_'):]
                     return f"{tag_type}['{short_field_name}']"
+            # Non-tag fields: look up the CUR1→CUR2 mapping for the SQL expression
+            if field in cur1to2_mapping:
+                return cur1to2_mapping[field]
+            # Field has no known CUR2 equivalent — return typed NULL
+            return empty.get(field_type.lower(), 'CAST(NULL AS varchar)')
 
 
     def column_surely_exist(self, field_to_expose):


NithinChandranR-AWS:
Closing -- this fix is superseded by #1503 which includes the same proxy fix plus additional PPA coalesce handling. Thank you.

Issue #1475, opened 2026-05-12
[Docs] - Account vs resource tagging dimensions: granular if available, high level if not


aristosvo:
Hi all,

I had a question which has been bothering me for a while and couldn't get clear from the documentation alone.

I have tags on account level for a `BusinessUnit`. I have sometimes tags on resources for `BusinessUnit` as well. Is it possible to have these merged, and use account level tag `BusinessUnit` only if resource tag `BusinessUnit` is not available for cost selection?

Account X: No account level tags
 -> Has resource XY which has `BusinessUnit=MonkeyTown` -> result: Resource cost is (only) selectable via `BusinessUnit=MonkeyTown` in Taxonomy Explorer and as Dimension

Account Y: Account level tag `BusinessUnit= MonkeyTown`
 -> Has resource YZ which has `BusinessUnit=ClumsyRabbit` -> result: Resource cost is (only) selectable via `BusinessUnit= ClumsyRabbit` in Taxonomy Explorer and as Dimension

Thanks a lot!

aristosvo:
I'm working on a setup to make merging of tags possible.

**Step 1**: Verify which tags are available in multiple dimensions (account, resource, IAM Principal)
**Step 2**: Show list of these tags and over which dimensions
**Step 3**: Select the tags to be merged
**Step 4**: Define merge strategy (default: resource tags over any other)
**Step 5**: Create the view etc (this is for me the hard part as it is a bit shattered and all over the place)

@yprikhodko Any advice on above mentioned approach?

yprikhodko:
The approach sounds directionally correct. I'm working on implementing this in cid-cmd. Thanks for submitting PR, I will review it soon

aristosvo:
Thanks!

I got the first implementation working, that's the good news.

The bad news is that I am no Python expert and the coding definitely needs some better grouping and organizing I'm afraid. Results on our tagging coverage look wonderful though 🚀

aristosvo:
Hi @yprikhodko!

Saw https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/pull/1526, and https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/pull/1524, which seem follow-ups of https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/pull/1510.

Is there any indication on when this might land and/or what is still necessary to land this?

[...]

PR #1510, opened 2026-06-24
feat: merge tags if they are the same

aristosvo:
*Issue #, if available:* #1475

*Description of changes:* Make merging tags on different levels possible. This is still a very early draft, I'll try to execute a first test round on our own setup.

### Known Issues
- [x] Merging tags: it seems like an account tag results in field account_tag_BusinessSolution vs tag_business_solution for normal tags. I need to fix this before I can continue in my own setup, working on it. Resulting example diff of the `tags_json`: 
   ```diff
   - , json_format(CAST(MAP_FROM_ENTRIES(ARRAY[ROW('tag_business_solution', resource_tags['user_business_solution']),ROW('iam_principal', line_item_iam_principal),ROW('account_tag_BusinessSolution', tags['accountTag/BusinessSolution'])]) AS JSON)) tags_json
   + , json_format(CAST(MAP_FROM_ENTRIES(ARRAY[ROW('tag_business_solution', resource_tags['user_business_solution']),ROW('iam_principal', line_item_iam_principal),ROW('account_tag_BusinessSolution', tags['accountTag/BusinessSolution']),ROW('merged_business_solution', COALESCE(NULLIF(resource_tags['user_business_solution'], ''), NULLIF(cost_category['business_solution'], ''), NULLIF(tags['accountTag/BusinessSolution'], '')))]) AS JSON)) tags_json
   ```
- [x] Merging tags: these tags are not replacing existing tags per se. It can be that users want to compare merged tags against their original counterparts
- [x] Resulting merged tags are _'null'_  ➡️ only the first view was updated due to input not used in the latter pre-seeded/non-interactive runs
- [ ] Tag shows as 'Merged \<Tag Title\>'. This might need a better title, but I don't know what and how to do it (yet)
- [ ] It definitely exposes some gaps in our own tagging setup 🤣 

It needs definitely some improvements on code quality and readability, first focus was on functionality. Results look good on our version of the dashboards!

### DIFF
--- a/cid/common.py
+++ b/cid/common.py
@@ -2293,6 +2293,28 @@ class Cid():
             logger.debug(f'cur.tag_fields_by_dimension = {merge_map}')
             if merge_map and selected is None:
                 keys_to_merge, merge_strategy = self._resolve_merge_config(param_name, merge_map)
+            elif merge_map:
+                # In pre-seeded/non-interactive runs selected tags are already set,
+                # so reuse merge parameters without prompting.
+                merge_strategy = get_parameters().get(f'{param_name}-merge-strategy', 'resource_first')
+
+                raw_keys = (
+                    get_parameters().get(f'{param_name}-merges')
+                    or get_parameters().get(f'{param_name}-merge-keys')
+                    or []
+                )
+                if isinstance(raw_keys, str):
+                    keys_to_merge = {k for k in raw_keys.split(',') if k}
+                else:
+                    keys_to_merge = set(raw_keys)
+
+                # Also infer merge keys from explicit merged_* selections.
+                for name in (selected or []):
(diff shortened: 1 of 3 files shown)


NithinChandranR-AWS (commented):
Thanks for tackling this! The direction is right. A few observations on the implementation: the two normalisation functions (_bare_key and _tag_to_name) diverge over time and the duplicate tag_fields_by_dimension property definition could cause confusion. I'm working on a companion implementation using a self-contained TagMerger class that consolidates this into a single normalise_key() method — will open a separate PR referencing this one and issue #1475. Happy to collaborate further. Thank you for your time and help.

aristosvo:
Closed in favor of https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework/pull/1526

PR #1517, opened 2026-07-10
fix(cfn): Use ResourcePrefix parameter in cid-cfn.yml

### LINKED ISSUE #1045: CFN Template cid-cfn.yml doesn't account for ResourcePrefix
The aggregation template allows the customer to specify the parameter [ResourcePrefix](https://github.com/aws-samples/aws-cudos-framework-deployment/blob/748775fbe2add5d9cfbf7e00f1c360216ec256ad/cfn-templates/data-exports-aggregation.yaml#L57) - which we used to avoid some collisions with other deployments.

Unfortunately the main dashboards template doesn't account for this in the bucket name (at least - maybe other places). [Here](https://github.com/aws-samples/aws-cudos-framework-deployment/blob/748775fbe2add5d9cfbf7e00f1c360216ec256ad/cfn-templates/cid-cfn.yml#L1346) for example the bucket is hard coded with a cid- prefix - 

```
                Resource:
                  - !Sub arn:${AWS::Partition}:s3:::cid-${AWS::AccountId}-data-exports # prefix for data-exports hardcoded here
```

For the old 1.0 configuration the bucket name could be passed in as a parameter to handle this difference in naming, but with the above this isn't possible without directly changing the template.

Other resources in the initial aggregator template also use this resource prefix for naming e.g. CIDDatabase (AWS::Glue::Database) so I'm also concerned they won't be picked up correctly if the 2nd template isn't aware of the prefix used.

Apologies if I missed something.

NithinChandranR-AWS:
## Issue

Fixes #1045

## Description

`cid-cfn.yml` hardcodes the `cid` prefix in 4 resource ARN references (2 S3 bucket ARNs, 2 Glue database/table ARNs). When a customer deploys `data-exports-aggregation.yaml` with a non-default `ResourcePrefix` value, the dashboard stack's IAM policy references the wrong bucket and Glue database.

## Changes

- Added `ResourcePrefix` parameter to `cid-cfn.yml` (default: `cid`, matching the data-exports-aggregation template)
- Replaced hardcoded S3 bucket ARN references with `${ResourcePrefix}-${AWS::AccountId}-data-exports`
- Replaced hardcoded Glue database/table ARN references using `!Join ['_', !Split ['-', !Sub '${ResourcePrefix}_data_export']]` to match the naming convention in `data-exports-aggregation.yaml`

## Backward Compatibility

Default value is `cid` — existing deployments are unaffected. Only customers who explicitly set a different ResourcePrefix in their data-exports stack need to pass the same value here.

### DIFF
--- a/cfn-templates/cid-cfn.yml
+++ b/cfn-templates/cid-cfn.yml
@@ -29,6 +29,7 @@ Metadata:
       - Label:
           default: 'Technical Parameters. Please do not change.'
         Parameters:
+          - ResourcePrefix
           - AthenaWorkgroup
           - AthenaQueryResultsBucket
           - DatabaseName
@@ -90,6 +91,8 @@ Metadata:
         default: "QuickSight DataSet Refresh Schedule [DEPRECATED]"
       ShareDashboard:
         default: "Share Dashboards"
+      ResourcePrefix:
+        default: "Resource Prefix - Must match the ResourcePrefix used in data-exports-aggregation stack"
       LambdaLayerBucketPrefix:
         default: "LambdaLayerBucketPrefix - Please do not change"
       GlueDataCatalog:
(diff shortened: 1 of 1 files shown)


NithinChandranR-AWS:
Hi, gentle nudge on this PR — it's been quiet for a while and is ready for review whenever you have bandwidth. Thanks!

PR #1518, opened 2026-07-10
fix(quicksight): Detect identity region via list_namespaces API

### LINKED ISSUE #943: Use api to get default region
Today we get identity region of QS via try and error message

This can be replaced with list_namespaces procedure

```
 list(qs.get_paginator('list_namespaces').paginate(AwsAccountId=account_id).search("Namespaces[? Name=='default'].CapacityRegion"))[0]
Out[9]: 'us-east-1'
```

NithinChandranR-AWS:
## Issue

Fixes #943

## Description

Replaces the fragile try/error approach for detecting the QuickSight identity region with the `list_namespaces` API, which returns the `CapacityRegion` directly.

### Before (current behavior)
1. Call `describe_user` in the current region
2. If it raises `AccessDeniedException`, parse the error message with a regex to extract the identity region
3. This is brittle — the error message format could change without notice

### After (this PR)
1. Call `list_namespaces` and query `Namespaces[? Name=='default'].CapacityRegion`
2. Returns the identity region directly — no error parsing needed
3. Falls back to the old `describe_user` method if `list_namespaces` fails (e.g., missing permissions)

## Backward Compatibility

The old behavior is preserved as a fallback. If `list_namespaces` raises any exception (permissions, throttling, etc.), the code falls back to the original `describe_user` + regex approach.

### DIFF
--- a/cid/helpers/quicksight/__init__.py
+++ b/cid/helpers/quicksight/__init__.py
@@ -85,33 +85,47 @@ class QuickSight(CidBase):
     def identityRegion(self) -> str:
         if not self._identityRegion:
             try:
-                logger.info(f'Detecting QuickSight identity region, trying {self.region}')
-                username = get_parameters().get('quicksight-user', self.username)
-                parameters = {
-                    'AwsAccountId': self.account_id,
-                    'UserName': username,
-                    'Namespace': 'default'
-                }
-                self.client.describe_user(**parameters)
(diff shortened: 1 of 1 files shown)


NithinChandranR-AWS:
Hi, gentle nudge on this PR — it's been quiet for a while and is ready for review whenever you have bandwidth. Thanks!

